PR #9, opened 2023-12-13
[minisite] enable to use asciidoctor-java

rmannibucau:
@fpapon we need to discuss it since it is a breaking change for the "asciidoctor pool"


fpapon (approved):
Works well, tested on https://www.yupiik.io website generation.

rmannibucau:
@fpapon should we merge that and move the version to 1.2.0-SNAPSHOT since it breaks the minisite asciidoc API?

fpapon:
> @fpapon should we merge that and move the version to 1.2.0-SNAPSHOT since it breaks the minisite asciidoc API?

Yes, I think it make sense to move to 1.2.0!

rmannibucau:
@fpapon upgraded the version in the poms, once the build passes, let's merge :)

PR #13, opened 2024-02-13
[minisite] Update theme

fpapon:


rmannibucau:
Note: added breadcrumb and bottom page nav (inspired from mainstream doc solutions like docusaurus), can be worth checking it is compatible and complies to css var?

Issue #24, opened 2024-11-21
Asciidoctor rendering ignores image attributes


jtama:
**Describe the bug**
The only attribute honored is the "alt" attribute, and it may even be erroneous, as the no name attribute is the first to be looked up.
I think at least the following attributes should be honored : 

- Width
- Height
- roles

**To Reproduce**
```
image::whatever.png[Acme 300 200]
// Also
image::whatever.png[alt="Acme", 300 200]
```

**Actual behavior**
```
<img src="whatever.png" alt="Acme 300 200">
```

**Expected behavior**
`<img src="whatever.png" alt="Acme" width=300 height=200">`

**Additional context**
Unlabelled attributes should also be ignored if the `alt` attribute is provided.

fpapon:
@jtama make sense, do you plan to provide a PR?

jtama:
I can do that.

jtama:
Done

fpapon:
@jtama merged, thanks for your contribution!

fpapon:
@jtama 1.2.5 has been released and is available on central.

jtama:
This should be update in quarkus-web-asciidoc on next dependabot run

PR #25, opened 2024-11-21
[asciidoc] Width and height attributes are propagated to rendered html.

jtama:
Unlabelled attribute is also only used id tha alt attribute isn't explicitly specified

Links #24


asciidoc-java/src/test/java/io/yupiik/asciidoc/renderer/html/AsciidoctorLikeHtmlRendererTest.java
```diff
@@ -601,6 +601,54 @@ void embeddedImage(@TempDir final Path work) throws IOException {
                 """.replace("$base64", base64), renderer.result());
     }
 
+    @Test
+    void imageWidthAndHeight() {
```
rmannibucau, line 605: guess both should use `assertRenderingContent` but PR looks good, kudos!

fpapon (approved):
LGTM!

PR #41, opened 2025-07-15
[asciidoc] Support for relfile attrs

### LINKED ISSUE #40: [asciidoc] add support for relfileprefix and relfilesuffix
This allows to make xref work when file are rendered in a static site way (foo.adoc -> foo/index.html)

in that case:
```
relfileprefix=../
relfilesuffix=/
```

Found the code here:
https://github.com/yupiik/tools-maven-plugin/blob/master/asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/html/AsciidoctorLikeHtmlRenderer.java#L880

I could try to provide the PR if you want?

ia3andy:
- Added support for `relfileprefix`, `relfilesuffix` and `outfilesuffix` (Fixes #40)
- Allow header without a title which is optional (https://docs.asciidoctor.org/asciidoc/latest/document/header/)


asciidoc-java/src/test/java/io/yupiik/asciidoc/parser/ParserTest.java
```diff
@@ -131,6 +131,13 @@ void parseHeader() {
         assertEquals(Map.of("attr-1", "v1", "attr-2", "v2"), header.attributes());
     }
 
+    @Test
+    void parseHeaderWithoutTitle() {
```
rmannibucau, line 135: Is this actually valid? I would have rather throw an exception if trying to parse a header without a header 🤔

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/html/AsciidoctorLikeHtmlRenderer.java
```diff
@@ -880,15 +880,18 @@ protected void visitToc(final Body body) {
 
     // todo: enhance
     protected void visitXref(final Macro element) {
+        final var outFileSuffix = configuration.getAttributes().getOrDefault("outfilesuffix", ".html");
```
rmannibucau, line 883: Shouldnt also page attributs be read (header if not null) cause config can be static for all pages

asciidoc-java/src/test/java/io/yupiik/asciidoc/parser/ParserTest.java
ia3andy: Have a look to the link I provided in the desc, https://docs.asciidoctor.org/asciidoc/latest/document/header/, the title is optional..

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/html/AsciidoctorLikeHtmlRenderer.java
ia3andy: TBH I don't know, I hesitated between both..

ia3andy: Any reason why for `:imagesdir:` we used the conf?

rmannibucau: tempted to say anywhere it should be a header or conf and was a legacy from the first original impl

fpapon:
@ia3andy thanks for your contribution! I will start the release process.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/html/AsciidoctorLikeHtmlRenderer.java
ia3andy: ok I let you figure it out in a follow up PR :)

PR #44, opened 2025-07-17
Some fixes to mimic asciidocj renderring

ia3andy:
- Allow to use include just after the title (it was considered as a authorLine)
- Allow to set `:imagesdir:` as part of the document attributes
- `:imagesdir:` automatically adds the trailing `/` if needed
- Make sure we stop reading header on blank line when parsing author and revision
- Make sure include is parsed even if there is not other attributes before (attributes.isEmpty check was before)
- Also read header attributes for relfile...

### DIFF
--- a/asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
+++ b/asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
@@ -100,6 +100,7 @@ public class Parser {
     private static final Pattern UNORDERED_LIST_PREFIX = Pattern.compile("^(?<wildcard>\\*+) .+");
     private static final Pattern UNORDERED_LIST2_PREFIX = Pattern.compile("^(?<wildcard>-+) .+");
     private static final Pattern ATTRIBUTE_DEFINITION = Pattern.compile("^:(?<name>[^\\n\\t:]+):( +(?<value>.+))? *$");
+    private static final Pattern MACRO = Pattern.compile("^[a-zA-Z0-9_+:.-]+::[^\\[]+\\[.*\\]\\s*$");
     private static final Pattern ATTRIBUTE_VALUE = Pattern.compile("\\{(?<name>[^ }]+)}");
     private static final List<String> LINK_PREFIXES = List.of("http://", "https://", "ftp://", "ftps://", "irc://", "file://", "mailto:");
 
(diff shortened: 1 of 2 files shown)


asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/html/AsciidoctorLikeHtmlRenderer.java
```diff
@@ -299,7 +297,6 @@ public void visitParagraph(final Paragraph element) {
     @Override
     public void visitHeader(final Header header) {
         if (header.attributes().get("notitle") == null &&
-                !Boolean.parseBoolean(configuration.getAttributes().getOrDefault("noheader", "false")) &&
```
rmannibucau, line 302: this one looks ackward to me and will have side effects so minimum would be to have a renderer toggle to restore this behavior IMHO

ia3andy:
For your information here is what I am testing against:
https://github.com/ia3andy/quarkus-roq/blob/asciidoc-up/roq-plugin/asciidoc/integration-tests/src/main/resources/content/guides/my-doc.adoc

asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
```diff
@@ -1452,10 +1464,7 @@ private Map<String, String> readAttributes(final Reader reader, final ContentRes
                             ofNullable(reader.nextLine()).map(String::strip).map(it -> ' ' + it).orElse("");
                 }
                 attributes.put(matcher.group("name"), value);
-            } else if (attributes.isEmpty()) {
-                reader.rewind();
-                break;
-            } else {
+            } else if (MACRO.matcher(line).matches()) {
```
rmannibucau, line 1467: not sure the regex is true since the macro doesn't have to start the line and if we go with a regex i'd like to have a quick test before (line.contains("::") for ex) since most of the time we'll not match it

ia3andy:
I have pretty much the same test with asciidocj, just I flattened the structure (until we fix the issue with nested includes)
https://github.com/ia3andy/quarkus-roq/tree/asciidoc-up/roq-plugin/asciidoc-jruby/integration-tests/src/main/resources/content/guides

ia3andy:
> @ia3andy are you sure for noheader? https://github.com/asciidoctor/asciidoctor/blob/main/lib/asciidoctor/converter/html5.rb#L180

Well this is what it does, maybe I am missing something..
https://github.com/ia3andy/quarkus-roq/blob/asciidoc-up/roq-plugin/asciidoc-jruby/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/runtime/AsciidoctorJConverter.java#L54

asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
ia3andy: That makes sense for the check, we could have `isMacro(line)`. The regex is generated by AI, we need to test it and improve it for sure.

ia3andy:
> > here are still issues in the rendering:
> 
> do you have the sources/adoc publicly available?

I've shared it above 👆

ia3andy:
> > https://github.com/ia3andy/quarkus-roq/blob/asciidoc-up/roq-plugin/asciidoc-jruby/runtime/src/main/java/io/quarkiverse/roq/plugin/asciidoctorj/runtime/AsciidoctorJConverter.java#L54
> 
> tested v3 and noheader skips the header, isnt it an issue cause you use a custom attribute and not the planned api? in any case makes > 6 years noheader skips it so I think we do not need to change that (except using attr instead of config.attributes where i fully agree)
> 
> will look the reste later

No sure what you mean by custom attribute? Did you test with asciidoctorj?

ia3andy:
Ok I tested and it shows the title because of showtitle:
```
asciidoctor guides/my-doc.adoc -e -s -a showtitle=true
```

ia3andy:
Which is what is documented AFAIK:

noheader | empty | No | Yes | Turns off header.
-- | -- | -- | -- | --
notitle | empty | No | Yes | Hides the doctitle in an embedded document. Mutually exclusive with the showtitle attribute.

noheader

empty

No

Yes

Turns off header.

notitle

empty

No

Yes

[Hides the doctitle in an embedded document](https://docs.asciidoctor.org/asciidoc/latest/document/title/#hide-or-show). Mutually exclusive with the showtitle attribute.

ia3andy:
@rmannibucau 
- for the BLOCK_MACRO pattern, could you have a look to it please? it seems block macro is always on a single line and at the start.
- for the `showtitle` I've also changed the implementation to show only if `showtitle` is true so it shouldn't break existing

rmannibucau:
FYI https://github.com/yupiik/tools-maven-plugin/pull/45 for the relative include

ia3andy:
Ok my mistake, I think I just mixed markdown and asciidoc always switching 😅

ia3andy:
All good then!

rmannibucau:
> Ok my mistake, I think I just mixed markdown and asciidoc always switching 

think asciidoctor enabled to support both and we did on some parts so still something we can enhance/fix IMHO

@ia3andy is it too much to ask you to rename BLOCK_MACRO to SIMPLISTIC_HEADER_BLOCK_MACRO to highlight it shouldn't be used anywhere else and is a light version of the actual parsing (ideally we fix it later merging it with the actual line parser)?

@fpapon overall ok for me mixed with https://github.com/yupiik/tools-maven-plugin/pull/45

ia3andy:
FYI I renamed it to HEADER_MACRO

ia3andy:
Hello @fpapon, could you trigger a release?
Have a good day!

fpapon:
> Hello @fpapon, could you trigger a release? Have a good day!

Yes, I will do it today

fpapon:
@ia3andy release 1.2.11 done and available on central

ia3andy:
> @ia3andy release 1.2.11 done and available on central

Thanks!!

PR #48, opened 2025-07-22
Fix 'data:' when using imagesdir

ia3andy:


ia3andy:
Wait for merging, test is not yet valid

PR #83, opened 2026-06-09
[minisite] llm support

rmannibucau:
can need some tuning/refinement but default should be functional assuming content is sufficient


fpapon:
Nice feature!
Thank you

PR #105, opened 2026-09-07
[fix] ensure PDF/Slide dependencies (requires) are better wired in as…

rmannibucau:
…ciidoctor instance


rmannibucau:
cc @fpapon

PR #111, opened 2026-09-08
[asciidoc] Fixes #50, render callouts as asciidoctor does when icons are enabled

### LINKED ISSUE #50: [asciidoc] Improve callouts rendering
Here is what I have with asciidocj:

Image

and the same with yupiik:

Image

Here is the html with asciidocj:
```html
<div class="sectionbody">
<div class="paragraph">
<p>The default directory structure is:</p>
</div>
<div class="listingblock">
<div class="content">
<pre class="highlightjs highlight"><code class="language-none hljs">my-site/
├── data/                                <i class="conum" data-value="1"></i><b>(1)</b>
│   ├── menu.yml
│   └── tags.yml
│
├── content/                             <i class="conum" data-value="2"></i><b>(2)</b>
│   ├── posts/                           <i class="conum" data-value="3"></i><b>(3)</b>
│   │   ├── 2024-10-14-roq-ssg/
│   │   │   ├── index.md
│   │   │   └── image.jpg
│   │   └── 2024-10-20-heart-roq.md
│   │
│   ├── roq-page.md                      <i class="conum" data-value="4"></i><b>(4)</b>
│   └── index.html                       <i class="conum" data-value="5"></i><b>(5)</b>
│
├── public/                              <i class="conum" data-value="6"></i><b>(6)</b>
│   └── style.css
│
└── templates/                           <i class="conum" data-value="7"></i><b>(7)</b>
    ├── partials/                        <i class="conum" data-value="8"></i><b>(8)</b>
    │   ├── head.html
    │   └── pagination.html
    │
    └── layouts/                         <i class="conum" data-value="9"></i><b>(9)</b>
        ├── base.html
        ├── page.html
        └── post.html
└── pom.xml                              <i class="conum" data-value="10"></i><b>(10)</b></code></pre>
</div>
</div>
<div class="colist arabic">
<table>
<tbody><tr>
<td><i class="conum" data-value="1"></i><b>1</b></td>
<td><strong>Roq Data Files</strong>
The <code>data/</code> directory contains data files like <code>menu.yml</code> and <code>tags.yml</code>. These files can hold structured data used across the site.</td>
</tr>
<tr>
<td><i class="conum" data-value="2"></i><b>2</b></td>
<td><strong>Content Files</strong>
The <code>content/</code> directory is where all content of the site that will be generated as pages resides (.html, .md, .adoc*, .json, .xml, …​)</td>
</tr>
<tr>
<td><i class="conum" data-value="3"></i><b>3</b></td>
<td><strong>Collections</strong>
The <code>posts/</code> directory is the default collection in Roq, it is optional. It holds content files for blog posts or similar structured documents. You can configure multiples collections (recipes, events, …​).</td>
</tr>
<tr>
<td><i class="conum" data-value="4"></i><b>4</b></td>
<td><strong>Additional pages</strong>
You may provide additional pages files like <code>roq-page.md</code> outside a collection. You can also use sub-directories which will be part of the resulting path.</td>
</tr>
<tr>
<td><i class="conum" data-value="5"></i><b>5</b></td>
<td><strong>Index File</strong>
The <code>index.html</code> file is required and serves as the homepage. It provides site-wide data using FrontMatter.</td>
</tr>
<tr>
<td><i class="conum" data-value="6"></i><b>6</b></td>
<td><strong>Static Files</strong>
The <code>public/</code> directory holds static files, such as images, CSS, or JavaScript files. These files will not be processed.</td>
</tr>
<tr>
<td><i class="conum" data-value="7"></i><b>7</b></td>
<td><strong>Templates</strong>
The <code>templates/</code> directory is optional as templates can be provided by a theme. It contains Qute templates for partials and layouts.</td>
</tr>
<tr>
<td><i class="conum" data-value="8"></i><b>8</b></td>
<td><strong>Qute Partials</strong>
The <code>partials/</code> directory contains reusable Qute template fragments, such as <code>head.html</code> and <code>pagination.html</code>, which can be included in layouts.</td>
</tr>
<tr>
<td><i class="conum" data-value="9"></i><b>9</b></td>
<td><strong>Layouts</strong>
The <code>layouts/</code> directory defines the structure for pages and documents. For example:
<div class="ulist">
<ul>
<li>
<p><code>base.html</code> is the main layout.</p>
</li>
<li>
<p><code>page.html</code> and <code>post.html</code> are specific layouts for pages and posts.</p>
</li>
</ul>
</div></td>
</tr>
<tr>
<td><i class="conum" data-value="10"></i><b>10</b></td>
<td><strong>Build files</strong>
The build file such as a <code>pom.xml</code> is needed to configure the build. It contains dependencies for your site such as theme and plugins.</td>
</tr>
</tbody></table>
</div>
</div>
```

rolfedh:
Reworked after the two reviews: the model carries the callout positions, the renderer no longer parses the code text, and the record has no redundant component.

- `Code` is now `(value, options, inline, lineCallOuts)`: `lineCallOuts` holds one list per line of `value` with the callouts whose markers ended that line (empty lists for lines without one; empty when the block has no callout), and the parser removes the `<N>` markers from the code instead of rewriting them to `(N)`, so `value` is the plain code. `callOuts()` stays as a derived accessor (each callout once, in order of appearance) for the callout list and for compatibility; the library reads `lineCallOuts` itself and materializes the flat list once. The old constructor is gone rather than kept as one that cannot carry positions.
- The renderer writes the markers from `lineCallOuts` while emitting the lines, so a decorating visitor does no extra work. With `:icons:` set the marker is `<i class="conum" data-value="N"></i><b>(N)</b>` and the callout list is a table, which is what the asciidoctor stylesheet styles (`.conum[data-value]`, the difference in the screenshots of #50); without the attribute the output is byte-identical to before. `hightlight-callouts=false` prints `(N)` as text, as before.
- One protected method, `callOutMarker(icons, number, inList)`, covers the code and the list. `highlightCallOuts` is kept and deprecated.
- Several markers on one line come for free from the model (`a=b <1><2>`), so this overlaps with #112's hunk in `parseWithCallouts`; whichever lands second needs a trivial rebase.

Tests: `codeWithCallout` asserts the stripped code and the positions; `calloutWithIcons`, `severalCalloutMarkersOnTheSameLine`; the existing `callout` and `callouts` renderer tests pass unchanged. Today's quarkus.io guide sources: 317 of 327 parse with this branch alone, the rest being the shapes #117 covers.

Fixes #50.

### DIFF
--- a/asciidoc-java/src/main/java/io/yupiik/asciidoc/model/Code.java
+++ b/asciidoc-java/src/main/java/io/yupiik/asciidoc/model/Code.java
@@ -15,10 +15,29 @@
  */
 package io.yupiik.asciidoc.model;
 
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
-public record Code(String value, List<CallOut> callOuts, Map<String, String> options, boolean inline) implements Element {
+/**
+ * A code (listing) block or an inline code span.
+ *
+ * @param value        the code, callout markers removed.
+ * @param options      the block options.
+ * @param inline       true for an inline code span.
+ * @param lineCallOuts one entry per line of {@code value}, listing the callouts whose markers ended that line
+ *                     (empty lists for lines without a marker); empty when the block has no callout.
+ */
+public record Code(String value, Map<String, String> options, boolean inline,
+                   List<List<CallOut>> lineCallOuts) implements Element {
+    /**
+     * @return the callouts of the block in the order their markers appear in the code, each one once even when its
+     * marker sits on several lines. Computed from {@link #lineCallOuts()} on each call.
+     */
+    public List<CallOut> callOuts() {
+        return lineCallOuts.stream().flatMap(Collection::stream).distinct().toList();
+    }
+
     @Override
     public ElementType type() {
         return ElementType.CODE;
--- a/asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/html/AsciidoctorLikeHtmlRenderer.java
+++ b/asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/html/AsciidoctorLikeHtmlRenderer.java
@@ -988,16 +988,25 @@ public class AsciidoctorLikeHtmlRenderer implements Visitor<String> {
                 builder.append(" data-linenums=\"true\"");
             }
             builder.append(">");
-            var html = escape(element.value());
-            if (linenums) {
-                final var lines = html.split("\n");
-                final var numbered = new StringBuilder();
-                for (int i = 0; i < lines.length; i++) {
-                    numbered.append("<span class=\"linenums\">").append(i + 1).append("</span>").append(lines[i]).append('\n');
+            // the parser removed the markers from the code and says which callouts ended each line
+            final var highlight = !"false".equalsIgnoreCase(element.options().get("hightlight-callouts"));
+            final var icons = icons();
+            final var lines = element.value().split("\n", linenums ? 0 : -1);
+            for (int i = 0; i < lines.length; i++) {
+                if (linenums) {
+                    builder.append("<span class=\"linenums\">").append(i + 1).append("</span>");
                 }
-                html = numbered.toString();
+                builder.append(escape(lines[i]));
+                if (i < element.lineCallOuts().size() && !element.lineCallOuts().get(i).isEmpty()) {
+                    builder.append(' '); // markers sat at the end of the line, `a=b <1><2>` in the source
+                    for (final var callOut : element.lineCallOuts().get(i)) {
+                        builder.append(highlight ? callOutMarker(icons, callOut.number(), false) : "(" + callOut.number() + ")");
+                    }
+                }
+                if (linenums || i < lines.length - 1) {
+                    builder.append('\n');
+                }
             }
-            builder.append("false".equalsIgnoreCase(element.options().get("hightlight-callouts")) ? html : highlightCallOuts(element.callOuts(), html));
             builder.append("</code></pre>\n </div>\n </div>\n");
         } else {
             final var nowrap = element.options().containsKey("nowrap-option") || state.nowrap;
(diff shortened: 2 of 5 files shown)


asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/html/AsciidoctorLikeHtmlRenderer.java
```diff
@@ -1925,11 +1948,46 @@ protected String highlightCallOuts(final List<CallOut> callOuts, final String va
...
+    }
+
+    protected String callOutListMarker(final String icons, final int number) {
+        if ("font".equals(icons)) {
+            return "<i class=\"conum\" data-value=\"" + number + "\"></i><b>" + number + "</b>";
+        }
+        return callOutImage(icons, number);
+    }
+
+    protected String callOutImage(final String icons, final int number) {
```
rmannibucau: not against but then do callOutIcon (else collapse both but no reason one is more extensible than the other no?)

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/html/AsciidoctorLikeHtmlRenderer.java
```diff
@@ -1925,11 +1948,46 @@ protected String highlightCallOuts(final List<CallOut> callOuts, final String va
         if (callOuts.isEmpty()) {
             return value;
         }
-        var out = value;
-        for (int i = 1; i <= callOuts.size(); i++) {
-            out = out.replace(" (" + i + ")", " <b class=\"conum\">(" + i + ")</b>");
+        final var icons = icons();
+        final var matcher = CALLOUT_MARKERS.matcher(value);
+        final var out = new StringBuilder();
+        while (matcher.find()) { // a run can hold several markers, `a=b (1)(2)` for ex
```
rmannibucau: looks like it breaks the parser work, I'd prefer we adjust the model (trying to keep backward compat on API land with deprecated method for single callout), did you evaluate it? there is no reason there is a real parsing in the parser except for specific rendering things but there it is plain AST

rmannibucau (changes requested):
i'd like mainly another round around the parsing for callouts (i'm fine to enable List<List<CallOut>> or alike) but reparsing it sounds fishy, in particular when decorating the visitor you need to do the same work N > 1 times

rolfedh:
Reworked as you suggested, force-pushed as one commit. `Code` now carries `lineCallOuts`, one list of `CallOut` per line of the value, filled by the parser, which removes the markers from the code instead of rewriting them to `(N)`. The renderer writes the markers from that list while emitting the lines, so nothing parses the code text anymore, and a decorating visitor inherits the positions. `callOuts()` stays for the list; the four-argument constructor is kept and deprecated because it carries no positions. The description has the details.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/html/AsciidoctorLikeHtmlRenderer.java
rolfedh: Collapsed into one protected method, `callOutMarker(icons, number, inList)`; the list form only differs by the parentheses.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/html/AsciidoctorLikeHtmlRenderer.java
rolfedh: Agreed, that was the flaw. The parser now owns the positions (`Code.lineCallOuts`) and the renderer only reads them. One overlap to flag: this touches the same `parseWithCallouts` hunk as #112, so whichever lands second needs a trivial rebase, which I will do.

asciidoc-java/src/main/java/io/yupiik/asciidoc/model/Code.java
```diff
@@ -18,7 +18,27 @@
...
+ * A code (listing) block or an inline code span.
+ *
+ * @param value        the code, callout markers removed.
+ * @param callOuts     the callouts of the block, in the order of their list.
+ * @param options      the block options.
+ * @param inline       true for an inline code span.
+ * @param lineCallOuts one entry per line of {@code value}, listing the callouts whose markers ended that line
+ *                     (empty lists for lines without a marker); empty when the block has no callout.
+ */
+public record Code(String value, List<CallOut> callOuts, Map<String, String> options, boolean inline,
```
rmannibucau: we don't need `List<CallOut> callOuts` there anymore but we need a  `callOuts()` accessor for compat (`lineCallOuts.stream().flatMap(Collection::stream).toList()` is ok with `callOuts` being the list of list)

side note: ideally the lib callsites are updated to not use the new method instead of all one to avoid multiple materialzations

rmannibucau (changes requested):
can you simplify the new code api please? else LGTM

asciidoc-java/src/main/java/io/yupiik/asciidoc/model/Code.java
rolfedh: Done in the force-pushed revision: `Code` is `(value, options, inline, lineCallOuts)`, and `callOuts()` is derived from the lines (`flatMap` plus `distinct`, so a marker that sits on two lines with one entry lists it once). The library reads `lineCallOuts` itself; the renderer materializes the flat list once for the callout list. I dropped the old four-argument constructor rather than keep one that cannot carry positions; say if you would rather have it back as deprecated.

rolfedh:
Simplified as asked and force-pushed; details on the inline thread. Same note as before: this overlaps with #112 in `parseWithCallouts`, so whichever lands second gets a trivial rebase from me.

rmannibucau (approved):

fpapon:
@rolfedh thank you for your contribution! Can you resolve conflicts please?

rolfedh:
> @rolfedh thank you for your contribution! Can you resolve conflicts please?

You're welcome, @fpapon. I've resolved the conflict. Thank you!

Issue #110, opened 2026-09-08
[asciidoc] NullPointerException in Parser.parseLine when bold text starts with an ordered-list marker (`**1. Bold**`)


rolfedh:
### Version

`io.yupiik.maven:asciidoc-java` 1.2.16.

### Reproduction

```java
new Parser().parse("**1. Bold** text.\n", new Parser.ParserContext(ContentResolver.of(Path.of("."))));
```

throws

```
java.lang.NullPointerException: Cannot invoke "io.yupiik.asciidoc.parser.internal.Reader.rewind()" because "reader" is null
    at io.yupiik.asciidoc.parser.Parser.parseLine(Parser.java:1174)
    at io.yupiik.asciidoc.parser.Parser.parseLine(Parser.java:1134)
    at io.yupiik.asciidoc.parser.Parser.addTextElements(Parser.java:2249)
    at io.yupiik.asciidoc.parser.Parser.parseLine(Parser.java:1278)
```

The trigger is bold text whose content starts with an ordered-list marker (`1. `). The same text without the number (`**Bold** text.`) parses, and the same number outside the bold markup (`1. Bold text.`) parses as a list. A section title in that form (`===== **1. Memory Usage Metrics (Every allocator)**`) fails the same way, which is how it was found.

### Expected

No exception: the bold span is inline text, so `1. ` inside it is not a list marker.

### Context

Found while converting the quarkus.io guides with asciidoc-java for quarkiverse/quarkus-roq#1105: `telemetry-micrometer.adoc` has five headings of the form `===== **1. ...**` and cannot be parsed at all.

Issue #106, opened 2026-09-08
[asciidoc] Several callout markers on one code line are rejected (Asciidoctor accepts `<1><2>`)


rolfedh:
### Version

`io.yupiik.maven:asciidoc-java` 1.2.16 (same on 1.2.14).

### Reproduction

```java
new Parser().parse("""
        [source,properties]
        ----
        a=b <1><2>
        ----
        <1> one
        <2> two
        """, new Parser.ParserContext(ContentResolver.of(Path.of("."))));
```

throws

```
java.lang.IllegalArgumentException: Invalid callout references (code markers don't match post-code callouts) in snippet: [Text[style=[], value=a=b <1><2> , ...
```

A marker number used on several lines with one callout entry (`int a; // <1>` and `int b; // <1>`, then `<1> both`) parses fine; only several markers on the same line trip the check.

### Expected

Asciidoctor accepts several callout markers on one line. The quarkus.io CDI reference guide has `quarkus.arc.exclude-types=org.acme.Foo,org.acme.*,Bar <1><2><3>` followed by three callout entries and renders with Asciidoctor.

### Context

Found while converting the quarkus.io guides with asciidoc-java for quarkiverse/quarkus-roq#1105 (a Markdown output for AI agents): 32 of 311 guide files fail to parse on this check alone, and the exception loses the whole document. Treating the markers positionally, or a lenient mode that keeps the document and drops the callout list, would help. Related to the rendering side in #50.

PR #117, opened 2026-09-08
[asciidoc] Parse callout lists after a paragraph and with attached blocks, never drop the document

rolfedh:
Reworked after the review: the detached-list support is gone, failure stays the default, tolerance is opt-in, and there is no logger.

Two things asciidoctor accepts in a callout list that the parser did not:

- An item can carry blocks attached with `+`, as for any list item; the continuation reader is the one the ordered and unordered lists use, `readContinuation`, which now takes its stop condition as a predicate (the next list marker there, the next `<n>` line here). The AsciiDoc docs on compound list items name callout lists explicitly ("callout and description lists also support compound list items"), and the Quarkus guides use it (`elasticsearch.adoc`, `hibernate-orm.adoc`). Item text can also wrap on a line starting with `<`, such as an xref.
- A list that does not match the markers still renders in asciidoctor, with a warning. The parser keeps throwing by default, as before; `:callout-mismatch: ignore`, as a document attribute or a parser attribute, opts into keeping the code with the callouts it read. The name follows asciidoctor's `attribute-missing`; happy to rename it.

A callout list separated from its block by a paragraph is not supported, on purpose: the docs place the list below the block with nothing between, and Quarkus moved its own lists for that reason (quarkusio/quarkus#56222, quarkusio/quarkus#56449).

On today's quarkus.io guide sources (327 files) this branch parses 319 by default (the rest are genuine mismatches in the sources, plus #110 which #116 fixed) and 326 with the attribute set.

Tests: `codeCalloutsWithAttachedBlocks`, `codeCalloutsMismatchFails`, `codeCalloutsMismatchIgnoredOnDemand`.

Relates to #106.

### DIFF
--- a/asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
+++ b/asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
@@ -498,11 +498,8 @@ public class Parser {
                     elements.add(parseSection(enclosingDocument, reader, options, resolver, attributes));
                 }
                 options = null;
-            } else if (Objects.equals("----", stripped)) {
-                elements.add(parseCodeBlock(enclosingDocument, reader, options, resolver, attributes, "----"));
-                options = null;
-            } else if (Objects.equals("```", stripped)) {
-                elements.add(parseCodeBlock(enclosingDocument, reader, options, resolver, attributes, "```"));
+            } else if (Objects.equals("----", stripped) || Objects.equals("```", stripped)) {
+                elements.add(parseCodeBlock(enclosingDocument, reader, options, resolver, attributes, stripped));
                 options = null;
             } else if (Objects.equals("--", stripped)) {
                 elements.add(parseOpenBlock(enclosingDocument, reader, options, resolver, attributes, "--"));
@@ -1066,7 +1063,21 @@ public class Parser {
             return new Code(subs(code, currentAttributes, substitutions), List.of(), codeOptions, false);
         }
 
-        final var callOuts = new ArrayList<CallOut>(contentWithCallouts.callOutReferences().size());
+        final var callOuts = parseCallOuts(enclosingDocument, reader, resolver, currentAttributes);
+        final var numbers = callOuts.stream().map(CallOut::number).toList();
+        if (!numbers.containsAll(contentWithCallouts.callOutReferences()) &&
+                // asciidoctor renders such a document with a warning; opt in to that with `:callout-mismatch: ignore`
+                !"ignore".equals(currentAttributes.getOrDefault("callout-mismatch", globalAttributes.get("callout-mismatch")))) {
+            throw new IllegalArgumentException("Invalid callout references (code markers don't match post-code callouts) in snippet:\n" + snippet);
+        }
+
+        return new Code(subs(contentWithCallouts.content(), currentAttributes, substitutions), callOuts, codeOptions, false);
+    }
+
+    private List<CallOut> parseCallOuts(final Path enclosingDocument, final Reader reader,
+                                        final ContentResolver resolver, final Map<String, String> currentAttributes) {
+        final var callOuts = new ArrayList<CallOut>();
+        String next;
         Matcher matcher;
         while ((next = reader.skipCommentsAndEmptyLines()) != null && (matcher = CALLOUT.matcher(next)).matches()) {
             int number;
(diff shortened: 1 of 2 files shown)


asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
```diff
@@ -100,6 +101,8 @@ public class Parser {
     private static final Revision NO_REVISION = new Revision("", "", "");
     private static final Header NO_HEADER = new Header("", List.of(), NO_REVISION, Map.of("authorcount", "0"));
 
+    private static final Logger LOGGER = Logger.getLogger(Parser.class.getName());
```
rmannibucau: for the record it was intentional there is no logger usage in the lib - except the Main but this one is particular, I also think it is sane to fail in the logging case - happy to have an attribute to ignore the error in "force" mode but default shouldn't let pass broken content IMHO

asciidoc-java/src/test/java/io/yupiik/asciidoc/parser/ParserTest.java
```diff
@@ -1185,6 +1185,118 @@ void codeIncludeNested(@TempDir final Path work) throws IOException {
                 body.children());
     }
 
+    @Test
+    void codeCalloutsAfterAParagraph() { // asciidoctor accepts a paragraph between the block and its callout list
```
rmannibucau: do you have the ref in the language doc it is acceptable, I rather read the opposite 🤔

asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
rolfedh: Agreed on both, and done in the force-pushed revision: the exception is the default again, exactly as before, and the tolerant path sits behind an attribute, `:callout-mismatch: ignore` (document or parser attribute). The name follows asciidoctor's `attribute-missing`; happy to rename it if you prefer "force" or something else. The logger is gone.

asciidoc-java/src/test/java/io/yupiik/asciidoc/parser/ParserTest.java
rolfedh: You are right, and I have removed it in the force-pushed revision. The callouts page places the list below the block with nothing between, and Quarkus moved its own lists for that reason (quarkusio/quarkus#56222 and quarkusio/quarkus#56449, after Guillaume called a separator incorrect in quarkusio/quarkus#56386). Asciidoctor's implementation tolerates the separation through its callout registry, which is why the shape existed in the guides, but that is an implementation tolerance rather than the language. What stays is the `+` continuation: the compound-list-items page says callout and description lists support them, and the Quarkus guides use it (elasticsearch.adoc, hibernate-orm.adoc).

asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
```diff
@@ -1016,26 +1027,67 @@ private Code parseCodeBlock(final Path enclosingDocument,
...
+        if (next != null) {
             reader.rewind();
         }
+        return callOuts;
+    }
 
-        if (callOuts.size() != contentWithCallouts.callOutReferences().size()) { // todo: enhance
-            throw new IllegalArgumentException("Invalid callout references (code markers don't match post-code callouts) in snippet:\n" + snippet);
+    // as for any list item, the text can wrap on the next lines and blocks can be attached with `+`
+    private void readCallOutContinuation(final Reader reader, final StringBuilder buffer) {
```
rmannibucau: did you try to consolidate readContinuation (maybe adding a Predicate<String> to match/stop at a line content)? I'm not super happy to have 2 ways to read continuation for future maintenance (even with AI ;)).
If too hard ok for me else I'd like it to be unified.

rmannibucau (commented):
LGTM, just the continuation point if it was thought

asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
rolfedh: Done: the two readers are one now, `readContinuation(reader, Predicate<String> stopsAt, buffer)`, with the list caller stopping on the next item marker and the callout caller on the next `<n>` line. The only behaviour that went with it is my tolerance for a blank line before the `+`, which none of the Quarkus guides use (28 attached blocks in today's sources, all with the `+` directly after the text) and which is the ancestor-attachment syntax anyway. Suite and corpus numbers unchanged.

rolfedh:
> LGTM, just the continuation point if it was thought

Thank you. Updated.

rmannibucau (approved):

rmannibucau (approved):

PR #119, opened 2026-09-09
[asciidoc] Fixes #118, don't read callout markers when the block drops that substitution

### LINKED ISSUE #118: [asciidoc] Callout markers are parsed even when the block disables them with subs="-callouts"
### Version

`io.yupiik.maven:asciidoc-java` 1.2.16.

### Reproduction

```java
final var ctx = new Parser.ParserContext(ContentResolver.of(Path.of(".")));
new Parser().parse("[source,java,subs=\"-callouts\"]\n----\nfoo(); // <1>\n----\n", ctx);
```

throws `IllegalStateException: Invalid state at line 5`, caused by `Invalid callout references (code markers don't match post-code callouts)`. With a `<1> ...` line after the block the parse succeeds but the block comes out as

```
Code[value=foo(); // (1)
, callOuts=[CallOut[number=1, ...]], options={language=java, subs=-callouts}, inline=false]
```

so the `<1>` was taken as a callout marker although the block disabled callouts.

### Expected

`subs="-callouts"` removes the callout substitution from the block (the block's default substitutions are `specialcharacters,callouts`), so asciidoctor renders `foo(); // <1>` as code text and any `<1> ...` line after the block as an ordinary paragraph. The same applies to a replacement list that does not name callouts, such as `subs="attributes"` (a list without `+`/`-` replaces the defaults; `verbatim` and `callouts` include the callout substitution, `normal` does not).

### Context

Found while converting the quarkus.io guides for quarkiverse/quarkus-roq#1105: `doc-create-tutorial.adoc` shows three code samples with `subs="-callouts"` where the `<n>` markers are deliberately part of the example text ("AsciiDoc callouts are shown as comments on some lines, and their numbering is not sequential"). It is the one guide left that the parser rejects on today's sources once #111, #112 and #117 are applied. Related to #106.

rolfedh:
A verbatim block substitutes `specialcharacters,callouts` by default, so `subs="-callouts"`, and any replacement list without callouts such as `subs="attributes"` or `subs="none"`, take the callout substitution away. `<1>` is then code text, and a `<1> ...` line after the block is an ordinary paragraph. The parser read the markers anyway: it rewrote them to `(1)` when a matching list happened to follow, and threw `Invalid callout references` otherwise.

`subs` is now resolved the way asciidoctor does it: an entry carrying a modifier (`+name`, `name+` or `-name`) changes the block defaults, a list without one replaces them, and the `none`, `normal`, `verbatim` and `specialchars` groups are expanded. The attribute substitution reads the same resolved set, with the block's own defaults (verbatim for listing, literal and source blocks, nothing for a passthrough block), so `subs="normal"` substitutes attributes too. The parser acts on `callouts` and `attributes` only; the other names are resolved so that the modifiers and the groups compute the right set, and change nothing, which a comment next to the constants states.

Tests: `codeWithCalloutsDisabledBySubs`, `codeWithCalloutsDisabledByASubsListWithoutThem`, `codeWithCalloutsKeptByAnIncrementalSubs` and `codeWithNormalSubs` in `ParserTest`, plus `calloutSubstitutionDisabled` in `AsciidoctorLikeHtmlRendererTest`, which asserts the markers render escaped, with no conum and no callout list.

Corpus: on today's quarkusio/quarkus guides (327 files), with #111, #112 and #117 applied, this takes doc-create-tutorial.adoc from rejected to parsed. That guide shows AsciiDoc callout syntax as example text, which is why it disables the substitution six times.

Fixes #118.

### DIFF
--- a/asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
+++ b/asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
@@ -58,6 +58,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -128,6 +129,18 @@ public class Parser {
     private static final Pattern REVISION_INFO = Pattern.compile("^(?:[^\\d{]*(.*?),)? *(?!:)(.*?)(?: *(?!^),?: *(.*))?$");
     private static final Pattern XML_TAG = Pattern.compile("<[^>]+>");
     private static final List<String> PREPROCESSOR_MACROS = List.of("ifdef", "ifndef", "ifeval", "endif", "include");
+    // substitutions as asciidoctor names them, a block selects them in its subs option (see resolveSubs()).
+    // the parser acts on two of them only: callouts (parseCodeBlock()) and attributes (subs()).
+    // specialcharacters, quotes, replacements, macros and post_replacements are resolved so that the +/- modifiers
+    // and the groups compute the right set, but change nothing here: a verbatim block is stored as written and
+    // the renderer always escapes it.
+    private static final Set<String> NO_SUBS = Set.of(); // passthrough block default
+    private static final Set<String> VERBATIM_SUBS = Set.of("specialcharacters", "callouts"); // listing, literal and source block default
+    private static final Map<String, List<String>> SUB_GROUPS = Map.of(
+            "none", List.of(),
+            "normal", List.of("specialcharacters", "quotes", "attributes", "replacements", "macros", "post_replacements"),
+            "verbatim", List.of("specialcharacters", "callouts"),
+            "specialchars", List.of("specialcharacters"));
 
     private final Map<String, String> globalAttributes;
 
@@ -584,8 +597,10 @@ public class Parser {
 
         final var text = content.toString();
         final var actualOpts = options == null ? Map.<String, String>of() : options;
+        // a literal block ("....") is verbatim, a passthrough block ("++++") has no substitution by default
+        final var substitutions = resolveSubs(actualOpts.get("subs"), "....".equals(marker) ? VERBATIM_SUBS : NO_SUBS);
         if (!text.contains("include::")) {
-            return new PassthroughBlock(subs(text, actualOpts), actualOpts);
+            return new PassthroughBlock(subs(text, actualOpts, substitutions), actualOpts);
         }
 
         final var filtered = Stream.of(text.split("\n"))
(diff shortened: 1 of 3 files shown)


asciidoc-java/src/test/java/io/yupiik/asciidoc/parser/ParserTest.java
```diff
@@ -1215,6 +1215,58 @@ String name (2)
...
+                List.of(
+                        new Code("foo(); // <1>\n", List.of(), Map.of("language", "java", "subs", "attributes"), false),
+                        new Text(List.of(), "<1> not a callout, this list replaces the default substitutions.", Map.of())),
+                body.children());
+    }
+
+    @Test
+    void codeWithCalloutsKeptByAnIncrementalSubs() {
+        final var body = new Parser().parseBody(new Reader(List.of("""
+                [source,java,subs="+attributes"]
```
rmannibucau, line 1255: not sure i get this one, shouldnt the code block contain an attribute and the callout be dropped there?

asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
```diff
@@ -128,6 +129,21 @@ public class Parser {
     private static final Pattern REVISION_INFO = Pattern.compile("^(?:[^\\d{]*(.*?),)? *(?!:)(.*?)(?: *(?!^),?: *(.*))?$");
     private static final Pattern XML_TAG = Pattern.compile("<[^>]+>");
     private static final List<String> PREPROCESSOR_MACROS = List.of("ifdef", "ifndef", "ifeval", "endif", "include");
+    private static final Set<String> VERBATIM_SUBS = Set.of("specialcharacters", "callouts");
```
rmannibucau: most of the values are not handled right? think we should either comment them to explicitly state it is not or support them

rmannibucau (changes requested):
think the attribute test can be enhanced to be more precise and I wonder if the new sub mapping is consistent with the code, can you please point me on the code I'm missing if so?

asciidoc-java/src/test/java/io/yupiik/asciidoc/parser/ParserTest.java
rolfedh: Two things in this one. The callout is kept on purpose: `+attributes` is the incremental form, it appends `attributes` to the block's default substitutions, which are `specialcharacters,callouts` for a verbatim block, so the callout substitution stays on. The callout goes away with `-callouts`, or with a list that has no modifier at all, such as `subs="attributes"`, because a plain list replaces the defaults; that case is `codeWithCalloutsDisabledByASubsListWithoutThem`. Reference: https://docs.asciidoctor.org/asciidoc/latest/subs/apply-subs-to-blocks/

You are right that the test proved nothing about the attribute side. It now has `{foo}` in the code and asserts both outcomes: the attribute is replaced and the callout is kept, so the incremental case shows the two switches at once.

asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
rolfedh: You are not missing any code, the mapping was ahead of it. The parser acts on two substitutions only: `callouts`, which the new check in `parseCodeBlock` reads, and `attributes`, which `subs()` reads. And `subs()` still looked at the raw string with `contains("attributes")`, so `subs="normal"` dropped the callouts through the resolver but did not substitute the attributes, which is exactly the inconsistency you sensed.

Fixed in 0476045: `subs()` now reads the resolved set too, with the block's own defaults (`specialcharacters,callouts` for a verbatim block, nothing for a passthrough block), so the composite names and the aliases behave the same on both switches, and `codeWithNormalSubs` covers `normal` (attributes replaced, callout markers left as text). The other names, `specialcharacters`, `quotes`, `replacements`, `macros` and `post_replacements`, are resolved only so that `+`, `-` and the composite names compute the right set; they change nothing here, because a verbatim block is stored as written and the renderer always escapes it. A comment next to the constants now says so, so the next reader does not have to ask. While there, the single-letter hints (`a`, `q`, ...) left the resolver: asciidoctor accepts them in the inline pass macro only, on a block they are invalid values, so the groups are now exactly `none`, `normal`, `verbatim` and `specialchars`.

rolfedh:
Thanks Romain, both points taken and answered in the threads. In short: the callout stays with `+attributes` because that form appends to the defaults, and the test now also proves the attribute replacement; and the mapping was indeed ahead of the code, so `subs()` now reads the same resolved set as the callout check, with a comment stating that only `callouts` and `attributes` change what the parser does. Still one commit, 0476045.

rmannibucau (approved):

rolfedh:
Closing: this change is already on master. François rebase-merged #121, which was stacked on this branch, so this commit landed as f949392 with the same patch and my authorship kept, and the #120 fix as 0c0fe43 on top of it. GitHub reports a conflict here only because the head SHA differs from the rewritten one. Thanks Romain for the review.

PR #126, opened 2026-09-11
[asciidoc] Fixes #102, read the inline form of ifdef/ifndef

### LINKED ISSUE #102: ifdef support in header processing does not handle inline conditionals
[Yupiik 1.2.16](https://github.com/yupiik/tools-maven-plugin/pull/87) handles block-form ifdef/ifndef/ifeval in readAttributes(), but silently drops the inline form (ifdef::attr[content] on a single line) and stops header parsing there. 

Block form works correctly (tested against 1.2.16):

```
= My Title
ifdef::env-github[]
:page-excluded: yes
endif::[]
:page-layout: post
```

→ page-layout is parsed correctly, page-excluded is absent (condition false). ✅

Inline form does not work:

```
= My Title
ifdef::env-github,env-browser,env-vscode[:imagesdir: ../assets/images/posts]
:page-layout: post
```
→ page-layout is null — parsing stops at the inline ifdef line. ❌

Also broken: inline form where the condition is true:

```
= My Title
:my-flag:
ifdef::my-flag[:page-set-by-inline: yes]
```

rolfedh:
Fixes #102

## What was broken

Only the block form of the conditional preprocessor macros, the one closed by a `endif::[]`, was known. The inline form `ifdef::attr[content]` carries its content in the brackets and has no `endif::[]`. It was mishandled in both places conditionals are read:

- In the header, `readAttributes()` tested `indexOf("[]")`, which is `-1` when the brackets hold anything, so the line fell through to the "plain line" case. The header ended there and every attribute below it was lost. This is the symptom reported in the issue.
- In the body, `readIfBlock()` looked for a `endif::[]` that never comes and swallowed the rest of the document into one conditional block, dropping the inline content itself. Same root cause, so it is fixed here too.

## What the fix does

- The brackets now tell the two forms apart: empty means a block, anything else means an inline whose content is the whole of the conditional.
- In the header the content is given back to the reader when the condition matches and dropped when it does not; either way header parsing goes on.
- In the body the content becomes the children of the `ConditionalBlock` and nothing more is read from the document.
- `ifeval` is excluded from the inline path since asciidoctor has no inline form for it.
- `ConditionalBlock.Ifdef` and `Ifndef` learned the multi-attribute conditions asciidoctor supports, `,` for an or and `+` for an and. The example in the issue (`ifdef::env-github,env-browser,env-vscode[...]`) relies on this; without it the header would parse but `imagesdir` would never be set.

## Verification

The three cases from the issue, from one reproducer against `master` and against this branch:

```
                                BEFORE (master)                AFTER
block form                      {page-layout=post}             {page-layout=post}
inline, condition false         {}                             {page-layout=post}
inline, env-github set          {}                             {imagesdir=../assets/images/posts, page-layout=post}
inline, condition true          {my-flag=}                     {page-set-by-inline=yes, my-flag=}
```

Checked against asciidoctor 2.0.26 rather than memory: for the header in the issue it renders `imagesdir` unset and `page-layout=post` with no attributes, and `imagesdir=../assets/images/posts` with `-a env-github`. The patched parser produces the same. The `,`/`+` truth table (including `ifndef` with both separators) matches asciidoctor for all four combinations of two attributes.

Tests: 314 pass in `asciidoc-java` (2 new tests: header cases and multi-attribute conditions in `ParserTest`, a body case in `ParserTest`, and an end-to-end `AsciidoctorLikeHtmlRendererTest` proving the inline content renders only when the attribute is set and never swallows what follows). 35 pass in `minisite-core`.

## Left out of scope, on purpose

Two adjacent issues surfaced while testing. Both predate this change and are shared with the block form, so they are filed separately.

- #127: `ifeval::[{version} > 1]` in a header is not recognized at all: `HEADER_MACRO` requires at least one character between `::` and `[`, and `ifeval` puts its whole condition inside the brackets, so the line is taken as the author line. Relaxing the regex also changes block-macro detection in the body parser.
- #128: a conditional block that directly adjoins paragraph text (block form or inline) renders without `<p>` and without a separator before the following line, because `visitParagraph`'s `addP` allowlist does not include `CONDITIONAL_BLOCK`. The block form `ifdef::foo[]` / `endif::[]` followed by a text line shows exactly the same output today.

### DIFF
--- a/asciidoc-java/src/main/java/io/yupiik/asciidoc/model/ConditionalBlock.java
+++ b/asciidoc-java/src/main/java/io/yupiik/asciidoc/model/ConditionalBlock.java
@@ -18,6 +18,8 @@ package io.yupiik.asciidoc.model;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Predicate;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 import static io.yupiik.asciidoc.model.Element.ElementType.CONDITIONAL_BLOCK;
 
@@ -25,6 +27,9 @@ public record ConditionalBlock(Predicate<Context> evaluator,
                                List<Element> children,
                                List<ConditionalBlock> elseBranches,
                                Map<String, String> options) implements Element {
+    private static final Pattern OR_SEPARATOR = Pattern.compile(",");
+    private static final Pattern AND_SEPARATOR = Pattern.compile("\\+");
+
     public ConditionalBlock(final Predicate<Context> evaluator,
                             final List<Element> children,
                             final Map<String, String> options) {
@@ -36,6 +41,24 @@ public record ConditionalBlock(Predicate<Context> evaluator,
         return CONDITIONAL_BLOCK;
     }
 
+    // as of asciidoctor a condition can list several attributes, "," being an "or" and "+" an "and",
+    // the two separators are not combinable so the first one found wins,
+    // an empty name ("a+" or "+" alone) is an undefined attribute
+    private static boolean isDefined(final String attribute, final Context context) {
+        if (attribute.indexOf(',') >= 0) {
+            return names(attribute, OR_SEPARATOR).anyMatch(it -> context.attribute(it) != null);
+        }
+        if (attribute.indexOf('+') >= 0) {
+            return names(attribute, AND_SEPARATOR).allMatch(it -> context.attribute(it) != null);
+        }
+        return context.attribute(attribute) != null;
+    }
+
+    private static Stream<String> names(final String attribute, final Pattern separator) {
+        return Stream.of(separator.split(attribute, -1)) // negative limit to keep the trailing empty names
+                .map(String::strip);
+    }
+
     @FunctionalInterface
     public interface Context {
         String attribute(String key);
@@ -44,14 +67,14 @@ public record ConditionalBlock(Predicate<Context> evaluator,
     public record Ifdef(String attribute) implements Predicate<Context> {
         @Override
         public boolean test(final Context context) {
-            return context.attribute(attribute) != null;
+            return isDefined(attribute, context);
         }
     }
 
     public record Ifndef(String attribute) implements Predicate<Context> {
         @Override
         public boolean test(final Context context) {
-            return context.attribute(attribute) == null;
+            return !isDefined(attribute, context);
         }
     }
 
--- a/asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
+++ b/asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
@@ -1634,13 +1634,17 @@ public class Parser {
                                 switch (macro.name()) {
                                     case "include" ->
                                             elements.addAll(doInclude(enclosingDocument, macro, resolver, currentAttributes, true));
-                                    case "ifdef" -> {
-                                        final var ifBlock = readIfBlock(reader);
-                                        elements.add(parseConditionalBlock("ifdef", macro.label(), ifBlock, enclosingDocument, resolver, currentAttributes, macro.options()));
-                                    }
-                                    case "ifndef" -> {
-                                        final var ifBlock = readIfBlock(reader);
-                                        elements.add(parseConditionalBlock("ifndef", macro.label(), ifBlock, enclosingDocument, resolver, currentAttributes, macro.options()));
+                                    case "ifdef", "ifndef" -> {
+                                        // ifdef::attr[content] is the inline form: the content sits in the brackets
+                                        // and there is no endif::[], so nothing more is read from the document
+                                        final var enclosed = line.substring(i + 1, end);
+                                        final var inline = !enclosed.isBlank();
+                                        final var ifBlock = inline ?
+                                                new IfBlock(List.of(enclosed), List.of(List.of(enclosed)), List.of()) :
+                                                readIfBlock(reader);
+                                        elements.add(parseConditionalBlock(
+                                                macro.name(), macro.label(), ifBlock, enclosingDocument, resolver, currentAttributes,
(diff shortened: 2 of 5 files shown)


asciidoc-java/src/main/java/io/yupiik/asciidoc/model/ConditionalBlock.java
```diff
@@ -36,6 +38,25 @@ public ElementType type() {
...
+        }
+        if (attribute.indexOf('+') >= 0) {
+            final var names = names(attribute, "+").toList();
+            return !names.isEmpty() && names.stream().allMatch(it -> context.attribute(it) != null);
+        }
+        return context.attribute(attribute) != null;
+    }
+
+    private static Stream<String> names(final String attribute, final String separator) {
+        return Stream.of(attribute.split(Pattern.quote(separator)))
```
rmannibucau: think we can pass the Pattern as separator to not recompile it multiple times

asciidoc-java/src/main/java/io/yupiik/asciidoc/model/ConditionalBlock.java
```diff
@@ -36,6 +38,25 @@ public ElementType type() {
...
     }
 
+    // as of asciidoctor a condition can list several attributes, "," being an "or" and "+" an "and",
+    // the two separators are not combinable so the first one found wins
+    private static boolean isDefined(final String attribute, final Context context) {
+        if (attribute.indexOf(',') >= 0) {
+            return names(attribute, ",").anyMatch(it -> context.attribute(it) != null);
+        }
+        if (attribute.indexOf('+') >= 0) {
+            final var names = names(attribute, "+").toList();
```
rmannibucau: guess we can avoid to materialize the list there

asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
```diff
@@ -2717,51 +2721,60 @@ private Map<String, String> readAttributes(final Path enclosingElement, final Re
...
-                                Map.of(), false);
-                        if ("ifdef".equals(macro.name()) || "ifndef".equals(macro.name()) || "ifeval".equals(macro.name())) {
-                            final var ifBlock = readIfBlock(reader);
-
-                            final var ctx = new ConditionalBlock.Context() {
-                                @Override
-                                public String attribute(final String key) {
-                                    return attributes.getOrDefault(key, globalAttributes.get(key));
+                final int sep = stripped.indexOf("::");
+                final int options = sep > 0 ? stripped.indexOf('[', sep) : -1;
```
rmannibucau: how can sep <0 in isBlockMacro if block?

asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
```diff
@@ -2717,51 +2721,60 @@ private Map<String, String> readAttributes(final Path enclosingElement, final Re
...
+                    // the conditioned content itself for the inline form of ifdef/ifndef which has no endif::[]
+                    final var enclosed = stripped.substring(options + 1, stripped.length() - 1);
+                    if ("ifdef".equals(macro.name()) || "ifndef".equals(macro.name()) || "ifeval".equals(macro.name())) {
+                        final var ctx = new ConditionalBlock.Context() {
+                            @Override
+                            public String attribute(final String key) {
+                                return attributes.getOrDefault(key, globalAttributes.get(key));
+                            }
+                        };
+                        boolean branchMatched = switch (macro.name()) {
```
rmannibucau: final

rmannibucau (changes requested):

asciidoc-java/src/main/java/io/yupiik/asciidoc/model/ConditionalBlock.java
rolfedh: Done: `OR_SEPARATOR` and `AND_SEPARATOR` are precompiled `Pattern` constants now and `names()` takes the `Pattern`.

asciidoc-java/src/main/java/io/yupiik/asciidoc/model/ConditionalBlock.java
rolfedh: Done. The list was only there to guard the empty case, and Asciidoctor does not special-case it: an empty name is just an undefined attribute, so `ifdef::a+[]` is false even with `a` set and `ifndef::+[]` is true (checked with asciidoctor 2.0.26). The split now keeps the trailing empty names (negative limit) and streams straight into `allMatch`, same result as Asciidoctor and no list. Added a test for the trailing separator.

asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
rolfedh: It can't. `isBlockMacro()` only lets through lines matching `HEADER_MACRO`, which requires the `::` (with at least one character before it) and the trailing `[...]`, so `sep > 0`, `options > 0` and `endsWith("]")` were all dead guards. Removed them: `sep` and `options` are plain `indexOf` calls now, the body lost one nesting level, and a comment points at the regex.

asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
rolfedh: Done.

rolfedh:
Thanks for the review. All four points are addressed in the amended commit a0f19a5 (kept as a single commit): precompiled patterns, no intermediate list (with Asciidoctor parity for empty names, test added), dead index guards removed in the header parsing, `final` on `branchMatched`. `mvn -pl asciidoc-java test`: 314 tests, 0 failures.

rmannibucau (approved):

PR #131, opened 2026-09-13
[asciidoc] Relates to #50, fail on a callout item without a marker in the code

rolfedh:
#117 made the parser accept a callout item whose marker is not in the code, and #111 then built the callout list from the code lines only, so such an item is dropped without a warning: a block with only a `<1>` marker followed by the items `<1>` and `<2>` lists one item. 1.2.16 rejected that document.

This keeps `Code` as it is and makes the mismatch check work in both directions: an item without a marker now fails like a marker without an item. `:callout-mismatch: ignore` still keeps the document, but, unlike #117, without that item, since `Code` has no place for it.

On the 279 deployed quarkus.io guides, 2 more fail by default than on master (21 instead of 19), and none fail with the attribute. Both are documentation tutorials whose example code contains a line like `include::_attributes.adoc[] <3>`: the parser expands it as an include, so the markers from 3 on disappear. That is a separate include bug, which I will fix in its own PR.


asciidoc-java/src/main/java/io/yupiik/asciidoc/model/Code.java
```diff
@@ -27,15 +27,18 @@
  * @param inline       true for an inline code span.
  * @param lineCallOuts one entry per line of {@code value}, listing the callouts whose markers ended that line
  *                     (empty lists for lines without a marker); empty when the block has no callout.
+ * @param callOuts     the callout list written after the block, in its order, including the items whose marker is
+ *                     not in the code.
  */
 public record Code(String value, Map<String, String> options, boolean inline,
```
rmannibucau, line 31: we explicitly decided to not go that path in previous round, IMHO the primary constructor should be the target designed (think data only) and convenient constructor can exist for backward compatibility but we never want both list IMHO since we moved to the list<list> to be generic, no? Am I missing something?

rmannibucau (changes requested):

asciidoc-java/src/main/java/io/yupiik/asciidoc/model/Code.java
rolfedh: You're right, this puts back the list you asked me to remove on #111. What I was trying to fix: the `List<List<CallOut>>` cannot hold a callout item whose marker is not in the code, for example when the marker sat on an `include::` line. There is no line to attach it to, so since #111 the item is dropped without a warning.

Two ways to keep a single list:

1. Treat it like the opposite case: fail with the mismatch error by default, as you asked on #117, and drop the item only with `:callout-mismatch: ignore`. No model change, and 1.2.16 rejected such a document too.
2. Keep one list with an extra entry after the last line for callouts without a marker.

I'd go with 1. OK for you?

rmannibucau: thinking out loud maybe a bit out of context but I think there is a line even for an include, it is just "virtual" but it is there no?

rolfedh: Good question. I checked with Asciidoctor 2.0.26. In a code block, `include::_attrs.adoc[] <3>` is not an include: text follows the `]`, so Asciidoctor keeps the line as code, with its marker. Our parser expands it, and it also drops every line after it, because `handleIncludes` returns after the first include. So the include case is a parser bug, not a model problem. I will fix it in a separate PR.

That leaves a description with no marker anywhere, like `<2>` below a block that has only `<1>`. Asciidoctor keeps it and logs "no callout found for <2>". The parser has no logger, so for that case I would keep option 1: fail by default, and keep the document with `:callout-mismatch: ignore`. OK for you?

rolfedh: I pushed option 1 (4fc7672) so you can see the code. `Code` is unchanged; only the mismatch check changed, in both directions. If you prefer to keep the description instead, it is a small change.

rmannibucau: yep, failing is the sanest default, an attribute as workaround is the backdoor if you have a need (since it is a bug it shouldn't and the adoc should be fixed to explicit the idea behind that but we already discussed it on the other PR ;))

rmannibucau (approved):

PR #132, opened 2026-09-13
[asciidoc] Read an include directive only as a whole line, and keep the lines after it

rolfedh:
An include directive is now read as asciidoctor's preprocessor reads it: `include::target[options]` must be the whole line (trailing spaces aside), with a target that is not empty and does not start or end with a space. One method, `includeDirective`, checks this without a regular expression, and blocks and text both use it.

- In a code, literal, passthrough or table cell block, `handleIncludes` expanded the first `include::` found anywhere, then dropped the rest of the block. So every line after an include was lost, and a line such as `include::_attributes.adoc[] <3>`, which is not a directive, was expanded and lost its callout marker (discussed on #131). The block is now read line by line.
- In text, the inline macro parsing expanded an include anywhere in a line. It is now text there, as in asciidoctor. A `\include::` in the middle of a line keeps its backslash; only an escaped whole line loses it, so the expectation of `escapedIncludeInText` (from #115) changed.
- An include inside a file that a code block includes is now expanded, relative to that file, as it already was in text.

Not changed here: an include inside a table is still not expanded, and `opts=optional` is still not recognized.

I compared 41 include positions with Asciidoctor 2.0.26: 35 give the same text now, 15 on master. On the 279 deployed quarkus.io guides, 3 guides change, all to what Asciidoctor renders: `doc-contribute-docs-howto` and `doc-create-tutorial` show their example headers instead of the whole attributes file, and `doc-reference` keeps `\include::_attributes.adoc[] // <3>` as written. Nothing else changes in the code or HTML of the guides, and the same 19 guides fail the callout check by default as on master.


asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
```diff
@@ -1146,38 +1148,44 @@ private List<Element> handleIncludes(final Path enclosingDocument,
...
-            if (start > 0 && content.charAt(start - 1) == '\\') { // escaped directive, asciidoctor drops the backslash and keeps the text
-                literal.append(content, from, start - 1).append(content, start, end + 1);
-                from = end + 1;
-                continue;
+        while (from < content.length()) {
+            final int newLine = content.indexOf('\n', from);
+            final int lineEnd = newLine < 0 ? content.length() : newLine;
+            final int next = newLine < 0 ? content.length() : newLine + 1;
+            final var line = content.substring(from, lineEnd);
+            final var directive = line.contains("include::") ? INCLUDE_DIRECTIVE.matcher(line) : null;
```
rmannibucau: not needed i think, we also tried hard to limit regex for includes so maybe it is worth checking there is no rest and we have a single parsing - thinking to macro case

rmannibucau (changes requested):
i'd like more tests with more edge cases cause it can change totally the way we do parse for goods but harnessing in a more complete doc and with more positions of the include in code blocks and doc are worth it - and a llm should be able to do it)

asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
rolfedh: Done in ace7a59: no regex anymore. `includeDirective` checks that the line starts with `include::`, ends with `]` (trailing spaces aside) and has a target without spaces around it. Code blocks and text both use it, so an include is parsed in one place, and the `include` case of the inline macro parsing is removed.

rolfedh:
I added tests with more edge cases (ace7a59).

`includesInACompleteDocument` has includes in a paragraph, a list continuation, an example block, a section with `leveloffset`, and a code block with callouts, `tag=`, an attribute in the target, trailing spaces and a nested include. It must give the same model as the same document with the includes written out, parsed with no file, so a line read as an include by mistake fails.

I also checked 41 include positions against Asciidoctor 2.0.26: 35 give the same text now, 15 on master.

One existing test changed: `escapedIncludeInText`, from my #115, expected `a line with \include::x[] in it` to lose its backslash. Asciidoctor keeps it; the backslash only goes for a whole line, now tested by `escapedIncludeLineInText`.

rmannibucau (approved):

Issue #130, opened 2026-09-12
[asciidoc] Contribute the Markdown renderer from quarkus-roq#1105


rolfedh:
### Version

`io.yupiik.maven:asciidoc-java` 1.2.16 and master.

### What already exists

I wrote a `MarkdownRenderer implements Visitor<String>` for quarkiverse/quarkus-roq#1105. It renders the model to GitHub-Flavored Markdown: ATX headings, fenced code with the language, pipe tables, admonitions as GitHub alerts, callouts as a numbered list after the code block, description lists, links and images. It is 815 lines with 22 unit tests, it imports only `io.yupiik.asciidoc.model` and `io.yupiik.asciidoc.renderer.Visitor`, and it compiles with `--release 17` against 1.2.16. I ran it over the 327 AsciiDoc files of quarkus.io and read the output beside the rendered HTML.

I would like to contribute it here as `io.yupiik.asciidoc.renderer.markdown.MarkdownRenderer`, beside `AsciidoctorLikeHtmlRenderer` and configured the same way through a `Configuration` object.

### Why here rather than in Roq

Two projects want this renderer. Roq publishes a Markdown copy of each page for AI agents. The Quarkus documentation RAG pipeline wants the same conversion to feed an embedding model (quarkusio/quarkus-documentation-rag#12), and it runs inside the Quarkus release build. Roq depends on Quarkus, so the RAG plugin cannot depend on Roq. `asciidoc-java` is the one library both projects already use. If the renderer does not live here, those two projects keep two copies of the same class and the copies drift apart.

### What I would change before opening the pull request

The Roq version re-parses callout markers with a regular expression, because the model did not record where they sat. #111 merged today, so `Code.lineCallOuts()` carries that now and the renderer can read the model instead. I would also drop its attribute regular expression wherever the model already gives me an `Attribute` element, and rewrite the tests on plain JUnit, since `asciidoc-java` does not use AssertJ.

### Question

Would you take it? If yes, do you prefer one pull request with the renderer and its tests, or a minimal renderer first and the rest as follow-ups?

rmannibucau:
Sounds good, 2 small questions/note: 1. Name should be about markfown flavors and not generic, 2. No way to fusion it in default html renderer?

rolfedh:
Thanks for the quick reply. Two proposals below; please tell me if either needs changing before I open the pull request.

[...]

PR #133, opened 2026-09-14
[asciidoc] Fixes #130, add GithubFlavoredMarkdownRenderer

rolfedh:
Adds `io.yupiik.asciidoc.renderer.markdown.GithubFlavoredMarkdownRenderer`, a `Visitor<String>` that writes the model as GitHub-Flavored Markdown: ATX headings, fenced code with the language, pipe tables, admonitions as GitHub alerts, callouts as a numbered list after the code, footnotes, anchors for ids, a table of contents, and `xref` targets built from `relfileprefix` and `relfilesuffix` as the HTML renderer does. It is configured through a nested `Configuration`, like `AsciidoctorLikeHtmlRenderer`, and a subclass's overrides also apply inside nested blocks.

It comes from quarkiverse/quarkus-roq#1105 and was checked on the 279 quarkus.io guides, with no unclosed fence or broken footnote in the output.

Known gaps, left for follow-ups:

- not rendered yet: table column alignment, author and revision lines, `:sectnums:`, `{counter:name}`, `[stem]` blocks
- a paragraph starting with raw HTML is not escaped
- the Maven plugin passes no attributes to a renderer it loads by class name, and the command line and the minisite only use the HTML renderer
- `IdGenerator` drops a separator that is not a word character, such as `-`, in both renderers
- a `[` inside a link label is not escaped, a bibliography entry does not show its `[ref]` label, and the `[qanda]`, `[horizontal]` and roman or alphabetic list styles are written as plain lists

How the parsing and the state are shared with the HTML renderer is discussed in the review.


_documentation/src/main/minisite/content/asciidoc-java.adoc
```diff
@@ -32,6 +32,42 @@ final var html = renderer.result(); <5>
...
+final var renderer = new GithubFlavoredMarkdownRenderer(new GithubFlavoredMarkdownRenderer.Configuration()
+    .setAttributes(Map.of("imagesdir", "/images"))); <1>
+renderer.visit(doc);
+final var markdown = renderer.result();
+----
+<.> Pass the attributes the renderer uses when the document does not define them, as for the HTML renderer.
+
+The Markdown renderer writes each construct as follows:
+
+* sections as ATX headings (`## Title`),
```
rmannibucau: not sure it was written manually or not but I think a table would be easier to read (I know tables are bad etc but still ;))

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
```diff
@@ -0,0 +1,1362 @@
...
+ * table cells) is rendered by the same renderer into a temporary buffer, so a subclass's overrides apply at any
+ * depth. When neither the document nor the configuration defines an attribute, its reference stays literal, as
+ * asciidoctor does with {@code attribute-missing=skip}. An unknown construct never throws, the renderer falls back to
+ * the element text.
+ * <p>
+ * A renderer can render several documents one after the other, {@link #visit(Document)} starts from a clean state.
+ * Like {@link io.yupiik.asciidoc.renderer.html.AsciidoctorLikeHtmlRenderer} it is not thread safe.
+ */
+public class GithubFlavoredMarkdownRenderer implements Visitor<String> {
+    private static final Set<String> ADMONITION_STYLES = Set.of("NOTE", "TIP", "IMPORTANT", "CAUTION", "WARNING");
```
rmannibucau: since 1-1 with the enum https://github.com/yupiik/tools-maven-plugin/blob/master/asciidoc-java/src/main/java/io/yupiik/asciidoc/model/Admonition.java#L28 why not using the num directly inline?

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
```diff
@@ -0,0 +1,1362 @@
...
+    private static final Set<String> ASCIIDOC_EXTENSIONS = Set.of(".adoc", ".asciidoc", ".ad", ".asc");
```
rmannibucau: asc? this is more a key, ultimately `adoc,asciidoc` should be the default and the value attribute (global or document scoped) configured no?

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
```diff
@@ -0,0 +1,1362 @@
...
+    private static final Pattern ATTRIBUTE_REFERENCE = Pattern.compile("\\{([a-zA-Z0-9_][a-zA-Z0-9_-]*)}");
+    private static final Pattern URL_SCHEME = Pattern.compile("^[a-zA-Z][a-zA-Z0-9+.-]*:");
```
rmannibucau: sounds too open no?

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
```diff
@@ -0,0 +1,1362 @@
...
+    private static final Pattern BLOCK_STYLE = Pattern.compile("[a-zA-Z][a-zA-Z0-9_+-]*");
```
rmannibucau: too open?

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
```diff
@@ -0,0 +1,1362 @@
...
+    private static final Pattern BACKTICK_RUN = Pattern.compile("`+");
```
rmannibucau: isnt it too easy to abuse and misparse? 🤔

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
```diff
@@ -0,0 +1,1362 @@
...
+        segments.add(current);
+        return segments;
+    }
+
+    private static boolean endsWithWhitespace(final Text text) {
+        final var value = text.value();
+        return value == null || value.isEmpty() || Character.isWhitespace(value.charAt(value.length() - 1));
+    }
+
+    private static boolean startsWithWhitespaceOrPunctuation(final Text text) {
```
rmannibucau: there are really a lot of parsing in this class, sounds like something is missed no?

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
```diff
@@ -0,0 +1,1362 @@
...
+    @Override
+    public String result() {
+        final var result = new StringBuilder(builder.toString().strip());
+        if (!state.footnotes.isEmpty()) {
+            result.append("\n\n");
+            for (final var footNote : state.footnotes) {
+                result.append("[^").append(footNote.label).append("]: ").append(footNote.text.strip()).append('\n');
+            }
+        }
+        return result.toString().strip() + "\n";
```
rmannibucau: +\n? (end with an empty line as a default rule?) - this is my practise

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
```diff
@@ -0,0 +1,1362 @@
...
+            case "image" -> {
+                blockAnchor(options);
+                builder.append(image(macro));
+                final var title = options.get("title");
+                if (title != null && !title.isBlank()) {
+                    builder.append("\n*").append(substitute(title).strip()).append('*');
+                }
+                builder.append("\n\n");
+            }
+            case "toc", "include" -> {
```
rmannibucau: why? 🤔 

toc can be generated and should include hit this renderer (else throwing an illegalargumentexception sounds sane)

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
```diff
@@ -0,0 +1,1362 @@
...
+                    "[" + target.replaceAll("([\\[\\]<>])", "\\\\$1") + "](" + destination(target) + ")") + closing;
+        }
+        return "[" + label + "](" + destination(url) + ")";
+    }
+
+    /**
+     * @return a link destination, in angle brackets when CommonMark would otherwise end it too early: on whitespace,
+     * on an angle bracket, or on a parenthesis without its pair.
+     */
+    protected static String destination(final String url) {
```
rmannibucau: personnally i'd remove most static from this class to enable now or later overrides/customizations, now for this particular method i'm not sure why we do parse that much, does the parser miss some parsing task?

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
```diff
@@ -0,0 +1,1362 @@
...
+            case "link" -> "[" + substitute(options.getOrDefault("", label)).strip() + "](" + destination(substitute(label).strip()) + ")";
+            case "xref" -> xref(macro, options);
+            case "mailto" -> "[" + options.getOrDefault("", label).strip() + "](" + destination("mailto:" + label.strip()) + ")";
+            case "kbd" -> keys(options.containsKey("opts") ? content + "," + options.get("opts") : content);
+            case "btn" -> "**" + content.strip() + "**";
+            case "menu" -> "**" + (label.isBlank() ?
+                    options.getOrDefault("", "").strip() :
+                    label.strip() + (options.get("") != null && !options.get("").isBlank() ? " > " + options.get("").strip() : "")) + "**";
+            case "pass" -> options.getOrDefault("", ""); // the label holds the substitutions, pass:q[...]
+            case "icon", "toc", "include", "indexterm" -> ""; // indexterm:[...] only feeds the index
```
rmannibucau: unsupported? some should issue some content (icon at least)

rmannibucau (changes requested):
put some comments/questoins inline, overall I find a lot of parsing and duplication so might be worth not integrating like that - or ensure we don't release it like that. Ideally all the parsing when needed (labels of macro) should be extracted in a _service_ like helper (`VisitorSibling`) any visitor can instantiate to handle that sharing the logic. The state management could also get a `VisitorState` (which is just a visitor handling the state and the actual renderer delegates everytime to maintain the state and is able to query it). With these two changes we should get something more robust for multiple different outputs cause there it is very error prone to change anything and we almost can't change the parser at all anymore without one regression (and LLM are super bad there :()

hope it makes sense, let me know if not

_documentation/src/main/minisite/content/asciidoc-java.adoc
rolfedh: Done in 407fd3d: the constructs and the attributes are two tables now.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
rolfedh: Done in 407fd3d: `admonitionLevel` loops on `Admonition.Level` and compares the exact name, as the parser and asciidoctor do. So `[note]` in lower case is no alert anymore, which was a listed follow-up.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
rolfedh: Right. Done in 407fd3d: `adoc,asciidoc` by default, and the `asciidoc-extensions` attribute changes it, from the document or the configuration. A link to another extension keeps the file name. For reference, asciidoctor itself only strips `.adoc` in `xref:`.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
rolfedh: Yes. Done in 407fd3d without a regex, with the rule of asciidoctor's `Helpers.uriish?`: a scheme has 2 characters at least, so `C:/images/a.png` is a path and gets `imagesdir`. It also stops at 32 characters, as a CommonMark autolink does.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
rolfedh: Yes, removed in 407fd3d. The fence language is the `language` option, else the style name from `styleName`, the method the admonitions use, except `source`, `listing` and `literal`. Before, `[listing]` gave a `listing` language.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
rolfedh: It only counts backticks, it does not parse the code: a Markdown fence must be longer than any run of backticks in the code, else a code line of three backticks closes the block. 407fd3d counts them with a loop instead of the regex, and inline code now uses the same count; it broke on a double backtick.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
rolfedh: Yes: the parser does not keep where a paragraph ends inside a delimited admonition or after a `+` continuation. The two paragraphs come as sibling `Text` elements, so the renderer guesses the boundary from the whitespace; the HTML renderer writes `First paragraphSecond paragraph` for the same input. I can open an issue for the parser, and remove this guess once each paragraph is a `Paragraph`.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
rolfedh: Yes, the output ends with one line feed, as a text file should. 407fd3d keeps it and builds the result with fewer copies.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
rolfedh: Done in 407fd3d. `toc::[]` writes the table of contents when `:toc: macro` is set, and `:toc:` or `:toc: preamble` place it at the top or after the preamble, as asciidoctor does. An include macro reaching the renderer throws an `IllegalArgumentException`.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
rolfedh: Done in 407fd3d: no static method is left. `destination` does not parse the URL, it escapes it for the Markdown syntax: CommonMark ends a link destination at a space or at an unbalanced parenthesis, so such a URL must be written between `<` and `>`.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
rolfedh: Done in 407fd3d: an icon is written as asciidoctor writes it without an icon font, `[lock]`, from the `alt` attribute or the icon name. `indexterm:[...]` stays empty, because asciidoctor shows nothing for it either; `indexterm2:[...]` shows its term.

rolfedh:
Thanks, it makes sense. The parsing in this class is of two kinds:

- escaping for the Markdown syntax (link destinations, fences, emphasis), which stays in the renderer;
- reading the model: ids and styles in the options, attribute references in titles, xref and image targets, macro labels, and the paragraph ends the parser loses.

I propose to move the second kind to a `VisitorSibling` that returns data only, and the state (document, attributes, parent elements, footnotes, id-to-title index) to a `VisitorState` visitor that the renderer calls from `visitElement`. In this pull request the Markdown renderer uses them; a next pull request moves the HTML renderer onto them, with its HTML output for the 279 quarkus.io guides unchanged. OK for you?

407fd3d answers the inline comments.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
```diff
@@ -0,0 +1,1662 @@
...
+
+    public GithubFlavoredMarkdownRenderer() {
+        this(new Configuration());
+    }
+
+    public GithubFlavoredMarkdownRenderer(final Configuration configuration) {
+        this.configuration = configuration;
+        this.state = new State();
+        this.context = key -> {
+            final var value = state.documentAttributes.get(key);
```
rmannibucau: (to not forget) can it be null? we never clearly defined if we enable to visitBody directly but not seeing why not, in partcular for md partial rendering - ok to not fix, just sharing the thought for future

rmannibucau:
yep, the output rendering should stay in the visitior for now (and until we get yet another md renderer)

side note that visitBody without visit() is an use case we have for the html renderer, not sure you want to handle it there but if so it can end with a NPE for attributes I think

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/VisitorSibling.java
```diff
@@ -0,0 +1,702 @@
...
+        }
+        return false;
+    }
+
+    /**
+     * @return the admonition level a block style names, {@code null} when it names none; as in asciidoctor and in the
+     * parser, the style must be written in upper case, {@code [note]} is no admonition.
+     */
+    public Admonition.Level admonitionLevel(final String style) {
+        for (final var level : Admonition.Level.values()) {
```
rmannibucau: `try { return Admonition.Level.valueOf(style);  } catch (final IllegalArgumentException e)  { return null; }` ?

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/VisitorSibling.java
```diff
@@ -0,0 +1,702 @@
...
+     *     <li>in a block container each inline element is a paragraph, except that a paragraph holding a line break is
+     *     given as its inline elements: around a line break, elements meeting on whitespace or punctuation are one
+     *     paragraph.</li>
+     * </ul>
+     *
+     * @param children       the sibling elements.
+     * @param paragraphLevel true for the children of a {@code Paragraph}, false for the children of a block container.
+     * @return the segments, in order.
+     */
+    public List<List<Element>> segments(final List<Element> children, final boolean paragraphLevel) {
```
rmannibucau: this one and its friends (until startsWithWhitespaceOrPunctuation) are very fishy and looks ultra specific, why do we need it for md and not html? 🤔 

shouldnt md also visit the structure instead of reflow it?

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/VisitorSibling.java
```diff
@@ -0,0 +1,702 @@
...
+            case PARAGRAPH -> ((Paragraph) element).children().stream().map(it -> plainText(it, context)).collect(joining());
+            default -> "";
+        };
+    }
+
+    /**
+     * @return the text of a section title on one line, as a table of contents or a link to the section shows it: its
+     * plain text without leading and trailing whitespace, each other run of whitespace read as one space.
+     */
+    public String titleText(final Element title, final ConditionalBlock.Context context) {
```
rmannibucau, line 264: for this kind of logic let's try to be optimistic - generally speaking not only this one - and not allocate out until we got proven we need to rewrite everything, if not (we can consider we always strip() which is very cheap - most cases) we just return text, means for well balanced spaces we return text.strip().

the common way to do it is to init out to null and lazy instantiate it and only flush it per segment (index) instead of per char

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/VisitorSibling.java
```diff
@@ -0,0 +1,702 @@
...
+        return explicit != null ? explicit : generatedId(title, context);
+    }
+
+    // ---------------------------------------------------------------------------------------------------- attributes
+
+    /**
+     * Replaces {@code {name}} references in a string the parser did not evaluate (block titles, image targets, ...);
+     * a reference to an attribute the context does not define stays literal.
+     */
+    public String substitute(final String text, final ConditionalBlock.Context context) {
```
rmannibucau, line 310: this one is a bit scary because it means we loose escaping no?

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/VisitorSibling.java
```diff
@@ -0,0 +1,702 @@
...
+        };
+    }
+
+    /**
+     * @param macroLevels the {@code levels} attribute of a {@code toc::[]} macro, {@code null} when there is none.
+     * @return how many section levels the table of contents lists, {@code toclevels} defaulting to 2 as in asciidoctor.
+     */
+    public int tocLevels(final String macroLevels, final ConditionalBlock.Context context) {
+        final var toclevels = context.attribute("toclevels");
+        final var levels = macroLevels != null && !macroLevels.isBlank() ? macroLevels : toclevels == null ? "2" : toclevels;
```
rmannibucau: i like inline coding but it can be wiser to use an "if" and return `2` when tocLevels is null?

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/VisitorSibling.java
```diff
@@ -0,0 +1,702 @@
...
+     * @return what a macro writes: its label, else its first positional attribute, as for {@code kbd:[Ctrl+C]} or
+     * {@code btn:[Save]}, which carry their content in the brackets.
+     */
+    public String content(final Macro macro) {
+        final var label = macro.label() == null ? "" : macro.label();
+        return label.isBlank() ? options(macro.options()).getOrDefault("", "") : label;
+    }
+
+    /**
+     * Splits the keys of a {@code kbd} macro as asciidoctor does: on the first of {@code ,} or {@code +} found after
```
rmannibucau, line 417: if specific to kbd maybe make it explicit in the name?

rmannibucau (changes requested):
a few remarks but I would push the exercise to try to reuse the sibling and potentially state for html renderer to validate the abstraction is not 100% specific to one renderer which would defeat the goal (if so it must sit in makrdown package but I don't think it is the case)

rmannibucau:
Spawned two llm reviews of this one to get another "opinion", will copy it but it is trasher than what I expected in the phrasing so just highlighting it is not my words ;). High level asked if we end up parsing in the rendering layer for bad reasons (ie parser bug).

I kind of agree on the outcome that we might have something broken between parser and renderer and think we should fix the parser.

(note: i assume you will give it to claude, no need to read both ;))

Here are their outputs:

FIRST:

# Ultra-critical review of yupiik/tools-maven-plugin#133 (asciidoc-java)

## The core architectural defect you flagged is real, and it is worse than a single method

`VisitorSibling.id()` is not an isolated case. The PR relocates/centralizes parsing that belongs in `Parser` into a renderer helper, and in doing so it codifies several parser leaks. The library even documents them in the code comments added by this PR:

- `VisitorSibling.java:64` — `// the parser stores [[id]] as [id]`
- `VisitorSibling.java:78` — `the parser also storing a [[id]] block anchor there, as [id]`
- `VisitorSibling.java:119` — `// the parser reads [%collapsible%open] as one "collapsible%open-option" key`
- `VisitorSiblingTest.java:61` — `// the parser keeps [[snippet]] as the style [snippet]`
- `VisitorSiblingTest.java:72` — `// the parser keeps one collapsible%open-option key`
- class javadoc `VisitorSibling.java:37-39` — "Reads what the model keeps as raw values ... a visitor parses"

Every one of these comments is an admission that the parser emits an unparsed/oddly-encoded value and the visitor reverse-engineers it. That is exactly the concern you raised.

---

## 1. `[[id]]` block anchor — confirmed, and it's a parser inconsistency, not just a parser gap

Path trace for a bare `[[snippet]]` line above a block:

1. `Parser.parseBody` main loop, `Parser.java:450`: `stripped.startsWith("[") && stripped.endsWith("]")` is true (a `[[id]]` line starts and ends with brackets), so it calls `parseOptions(stripped.substring(1, len-1))`, passing the inner string `[id]`.
2. `doParseOptions("[id]", "", true)` (`Parser.java:2906`): no `=`, no `,`, so `key=[id]`, value empty → `positional=["[id]"]`. The token does not start with `.`/`#`/`%`, so `Parser.java:2952` does `map.putIfAbsent("", "[id]")` — i.e. `[id]` is stored as the style.
3. The renderers never see an `id` key. `VisitorSibling.id()` de-brackets `[id]` back into `snippet` (`VisitorSibling.java:64-66`); `styleName()` special-cases a leading `[` to report "no style" (`VisitorSibling.java:82`).

Meanwhile the parser does correctly parse the same `[[id]]` syntax in the inline/text context — `Parser.newText()` (`Parser.java:2450-2468`) recognizes `[[id]]` and stores it under `id`. So the parser understands `[[id]]` at block level if you squint (it strips the outer pair) yet does not extract the id there, while in text it does. Same syntax, two contradictory outcomes, decided in the parser by context, and then patched in the visitor for one of them.

Compare with `[#id]`, which `Parser` handles correctly: leading `#` → `flushOption`/`doParseOptions` emit `id` key (`Parser.java:2545`, `2947`). So `[#snippet]` yields a clean `id` in the model, but `[[snippet]]` yields an opaque `[snippet]` style that the visitor must re-parse. The inconsistency is user-visible and drift-prone.

Verdict: this is not acceptable as designed. Parsing of `[[id]]` belongs in the parser. There is a second, different problem lurking inside `id()` too:

### 1b. `id()` also re-parses the `[NOTE#id.role%option]` block *style shorthand* — parsing the parser didn't do

I traced `[NOTE#note.important%collapsible]`:
- `Parser.parseOptions` (`Parser.java:2481-2499`) handles the `%collapsible` part into a `collapsible-option` key, but `doParseOptions` keeps the rest as the raw style string `NOTE#note.important` (positional → `putIfAbsent("", "NOTE#note.important")`).

So `VisitorSibling.id()` (`VisitorSibling.java:67-72`) re-decomposes `#id`.role` embedded in a style — parsing the parser left undone. `styleName()` (`VisitorSibling.java:80-88`) does the same `firstIndexOf(value, "#.%")` decomposition. **Both** `id()` and `styleName()` split on the *same* character set `#.%`, but with subtly different indexes (`value.indexOf('#')` vs `firstIndexOf("#.%")`), which is exactly the kind of near-duplicate logic that will drift. Only one form (`[#foobar]`` standalone) is parsed properly; every richer shorthand is only understood by the visitor.`

---

## 2. Corroborating defects: the same pattern in `hasOption`, `styleName`, `kbdKeys`

- `[%collapsible%open]` → key `"collapsible%open-option"`. The parser joins two options with `%` in a single key name and `VisitorSibling.hasOption()` re-splits it (`VisitorSibling.java:119-127`). `parseOptions` explicitly documents this (`Parser` `%` handling). This is another value that the parser encodes but does not parse.

- `styleName()` returns `""` for any style starting with `[` (`VisitorSibling.java:82`). This is heuristic discrimination ("starts with `[` means it's an id, not a style"). But `[[0]][1]`-style indexes, `[...]`-containing style names or any future `[...]`-leading option would be silently misclassified. Because the model stores `[id]` indistinguishably from a genuine style string, the visitor must mis-guess at the boundary — the root cause is that the parser never normalized the anchor.

- `kbdKeys()` (`VisitorSibling.java:422-447`) merges the macro `content` + `opts` and re-partitions on the first `,` or `+` after char 0. The comment ("the parser splits kbd:[Ctrl,Shift] at the comma") concedes the parser has already split the value, so the visitor is gluing two parser-split fragments back together and re-splitting differently. This is the same re-parse smell, plus it means the canonical tokenization is split across two layers.

---

## 3. Design-level concerns (beyond the specific bug)

- A second reader of the model. The PR introduces `VisitorSibling` + `VisitorState` as a parallel interpretation layer over the model, while the parser's contract stays "leave it raw, renderers interpret." That is a coherent stance, but the PR is inconsistent about it: some things are parsed (`[#id]`, `[%x]`, inline `[[id]]`) and others are deliberately not (`[[id]]` block, embedded `#`/`.`/`%` shorthands, `%`-joined option key, kbd). You either parse in the parser or in the sibling — not half of each.

- It hard-codes renderer decisions into a shared helper that subclasses are expected to override. The renderer javadocs say "a subclass of `VisitorSibling` changes a reading in both" — so the intended extension point is re-plumbing broken parsing, not rendering. A user who just wants a different id scheme is now forced to reason about the parser's raw encoding.

- Test coupling / locking in the wrong contract. `VisitorSiblingTest` asserts the parser's lossy encodings (`idAndStyleShorthands`): `[id]`-style, `collapsible%open` key, `NOTE#note.important`. Those tests bake the defect in. If the parser is fixed, these assertions must change; today they would actively block the correct fix.

- Scope/documentation. The PR body's "Known gaps" list is honest, but the structural gap ("[the]/[id]/opts shorthands are parsed in the visitor") is presented as normal, which masks its severity. `[[id]]` is the single most common way users anchor blocks; shipping a renderer that only works on this by rewriting `[[X]]` → `[X]` strings in the visitor is not a stable foundation to build `toc`, `xref`, and footnote resolution on.

---

## Concrete fix plan (what I would change)

1. Parser: normalize block-level `[[id]]` into an `id` option (primary fix).
   - In `Parser.parseBody` (or `parseOptions`), before the `[`/`]` block-attribute branch swallows `[[id]]`, detect a line matching `^\s*\[\[[^\]]+\]\]\s*$` and emit `Map.of("id", <inner>)` instead of treating the inner `[id]` as a style.
   - Keep the HTML/markdown behavior byte-identical for `[[snippet]]` after this change (both renderer tests pass unchanged, only the model differs — the sibling no longer needs the bracket branch).

2. Parser: decompose the style shorthand. Extend `doParseOptions` to split a bare positional style on `#`/`.`/`%` into `""`(style)/`id`/`role`/`*-option` — mirroring what line 2545 already does for a leading `#`. Then `VisitorSibling.id()` collapses to `return options == null ? null : options.get("id")` and `styleName()` to `options.get("")`; delete the `firstIndexOf`/shorthand branches in `id()`/`styleName()`.

3. Parser: stop joining `%`-options into one key (`[%collapsible%open]`). Split each `%`-segment into its own `open-option`/`collapsible-option` entry (the `parseOptions` `%` branch at `Parser.java:2481` already does this for the `source%opts` case; reuse it). Then delete the `%`-key re-splitting loop in `hasOption()`.

4. `kbdKeys`: have the parser keep the raw delimiters (don't pre-split on comma at the model boundary); let the single `kbdKeys` implementation be the only tokenizer. If that is too broad for this PR, at minimum add a comment + a parser-level unit asserting the two layers agree — but the cleaner fix is one owning layer.

5. Update tests to the corrected contract — `VisitorSiblingTest` and admin-affected tests in `GithubFlavoredMarkdownRendererTest` should assert `id(options) == "snippet"`, `styleName()=="NOTE"`, `hasOption=="collapsible"&&"open"` from parser-produced options, not from the lossy encodings.

6. Delete the now-dead heuristics in `VisitorSibling`: the `[`-prefix detection in `styleName()` (line 82) and the bracket-stripping in `id()` (line 64-66). `VisitorSibling` becomes a true reader (plain text, targets, footnotes, toc placement, uri checks — all legitimate renderer concerns), with no parsing left in it.

7. Re-run the 279-guide markdown check and the `AsciidoctorLikeHtmlRendererTest` after the parser change, to prove byte-identical output for existing docs while the model becomes correct.

### Risk / trade-offs
- The parser change is the riskiest because block-anchor behavior is consumed by the existing HTML renderer; the tie-breaker tests (`AsciidoctorLikeHtmlRendererTest`, which this PR also edits substantially at `+139/-3`) must stay green. Mitigate by making the parser emit the same final `id`/`""`/`role` values the sibling currently derives.
- Deleting sibling methods that subclasses may already override (it's a public, documented extension point in this PR) is a source-compat break — justified since the PR is not yet merged and the methods encode broken parsing.
- Keep the `[1]`-style (bibtex index) and any genuine `...`-leading styles working — the new parser decomposition must not convert those to ids.

### TL;DR
The `VisitorSibling` performing `[[id]]` (and `[id]`, `.role`, `%opts`, kbd) parsing is not acceptable as submitted. It's not one method — the PR introduces a consistent anti-pattern where the parser stores half-parsed, oddly-encoded values and a renderer-shared helper re-parses them, with tests that lock the lossy encodings in. The fix is to move the parsing into `Parser` (block `[[id]]`, style shorthand decomposition, `%`-option splitting), collapse the sibling to pure reading, and update the tests to the corrected model. Everything else in the renderer logic (admonitions, callouts, footnotes, toc, xref) is structurally reasonable; the blocker is the parser/sibling boundary.



SECOND:

Ultra-critical review of PR #133 (asciidoc-java scope)
Verdict: request changes — the architecture is inverted
The PR adds ~647 lines of VisitorSibling + ~1085 lines of GithubFlavoredMarkdownRenderer whose foundation is re-parsing syntax the parser failed to model. Your [[id]] example is not an isolated slip; it is the PR's design pattern.
Blocking: a visitor must not parse — the parser must
1. [[id]] block anchors. Parser.java:450-456 strips the outer brackets of [[id]] and parseOptions("[id]") dumps the raw string [id] into the style key. VisitorSibling.id() then strips those brackets back open. The parser must recognize a standalone [[id]] line as a block anchor and emit options.id. The test enshrines the bug: assertEquals("snippet", sibling.id(anchored)); // the parser keeps [[snippet]] as the style [snippet].
2. styleName() re-parses STYLE#id.role from the raw "" key, because doParseOptions (Parser.java:2943-2956) only handles leading ./#/% positionals. Style-shorthand decomposition is parser work.
3. hasOption() compensates for the parser mangling [%collapsible%open] into a single collapsible%open-option key (Parser.java:2949-2950, and the inline-% path at 2481-2498 explicitly skips tokens starting with %). The parser should emit one x-option per token; the visitor then deletes that compat branch.
4. substitute() re-implements attribute-reference substitution (with escape handling) that the parser already has (earlyAttributeReplacement, Parser.java:444) but does not apply to stored titles — so every renderer must re-substitute titles.
5. crossReference()/crossReferenceId()/documentName()/hasExtension()/isUri()/relativeFile()/documentPath() — the xref target grammar (file vs id vs fragment, extension rules) is syntax, parsed per-renderer in a visitor.
6. kbdKeys()/footnoteId()/footnoteText()/imageAlt()/defaultAlt()/imageTarget()/content() — macro attribute/content grammars re-parsed in the visitor layer.
7. descriptionListTitle() works around another parser bug (the :title: document attribute copied into list options) — and silently changes existing HTML output (lists that previously showed a bogus title stop showing it).
8. GFM segments()/splitParagraphRun()/splitBlockRun() guess paragraph boundaries from whitespace/punctuation of sibling Texts — structure discovery is the parser's job. Heuristics like "two unstyled texts meeting without whitespace are two paragraphs" will mis-split (**bold**text, attribute followed by punctuation).
9. HTML visitAsAdmonition copies the options map to inject the id parsed back out of [NOTE#id] — the renderer reconstructing what parseOptions should have produced.
Proof it is the wrong layer: if the parser parsed block anchors, style shorthands, %options, titles, and xref targets, VisitorSibling shrinks to nearly nothing.
Blocking: silent HTML behavior changes smuggled into a "new renderer" PR
- Image alt default relative.png → relative — existing tests (imagesDirs, imagesDirsTrailingSlash) were rewritten to match the new behavior. Red flag.
- visitXref: previously only .adoc targets were rewritten; now extension-less and any configured extension get relfileprefix/suffix, and display text comes from VisitorState titles instead of the xref catalog — different lookup semantics (catalog saw inline anchors too).
- Two footnote registries now exist (HTML State.footnotes and VisitorState.footnotes).
These output-visible changes must be separated, individually justified against asciidoctor, and not covered by edited-in-place tests.
Design issues
- Dead code deprecated in the same PR that adds it (extractPlainText, buildXrefCatalog, ensureXrefCatalog, State.xrefCatalog) — delete, don't deprecate.
- VisitorSibling imports renderer.html.IdGenerator — a supposedly format-agnostic class coupled to the HTML package; GFM emits "the id the HTML renderer would give". Id generation belongs to parser/model.
- VisitorState.visitBody clears sectionTitles/referencedIds/tocSections but not footnotes/labels/counters — the advertised "render a body alone" reuse leaks footnotes.
- link() strips a trailing > from the URL — the parser should not keep the closing bracket of <https://...> in the URL.
- PR bundles four changes: parser-gap workarounds + shared visitor infra + HTML refactor + new renderer. The order is wrong.
Plan (the required re-ordering)
1. Parser first (Parser.java + ParserTest):
- block anchor [[id]] line → options.id (sections, blocks, list items);
- decompose first positional Style#id.role%opt into style/id/role/x-option;
- normalize leading %a%b into multiple -option keys;
- substitute (or model) block-title attribute references so renderers drop substitute();
- parse the xref target into a model/parser-level structure (file/id/fragment).
2. Slim VisitorSibling to what genuinely remains; move id generation out of renderer.html; delete the [[id]]-as-[id] and collapsible%open-option compat branches and their enshrining tests.
3. HTML renderer: delegation-only diff; split output-visible changes into separate, justified commits/tests.
4. Delete deprecated dead code; fix VisitorState footnote clearing.
5. GFM renderer last, with segments() replaced by parser-guaranteed paragraph modeling (or the limitation documented in the parser, not guessed around).

rolfedh:
Thanks, I agree: the parser should give these values ready to use. The `[[id]]` line, the `NOTE#id.role` shorthand, `%collapsible%open`, attributes in block titles, `:title:` in description lists, `kbd:[Ctrl,Shift]` and paragraph ends are parser gaps; asciidoctor handles them in its parser. Two points are not right: the xref catalog methods are in 1.2.16, so they are deprecated, not dead code of this PR; and fixing `[[id]]` in the parser changes the HTML of 152 guides, since the HTML renderer ignores that id today.

My plan: one small parser pull request per gap first, like #136 and #137; #138 fixes `[[id]]`. When each one is merged, I remove the matching code from `VisitorSibling` here. Then this PR keeps only the readings that are not parsing, and the HTML renderer changes get their own pull request. OK for you?

rolfedh:
Status of this PR: I am not asking for a new review yet, because the code will still change.

1. When #138, #139 and #140 are merged, I rebase this PR and remove the matching readings from `VisitorSibling`: the `[[id]]` style, `descriptionListTitle()` and the title substitution.
2. I move the HTML renderer changes to a separate pull request, as I wrote above.
3. The paragraph groups depend on your answer on #141. After that, the Markdown renderer uses the groups from the parser, and `splitParagraphRun()` goes.

When this is done and CI passes, I will ask for your review again, with a list of what changed since your last review.

rolfedh:
@rmannibucau @fpapon, this pull request is not ready for a new review yet. [My plan above](https://github.com/yupiik/tools-maven-plugin/pull/133#issuecomment-5679556191) still applies. [#139](https://github.com/yupiik/tools-maven-plugin/pull/139) and [#140](https://github.com/yupiik/tools-maven-plugin/pull/140) are merged, so I now remove their readings from `VisitorSibling` and move the HTML renderer changes to a separate pull request. [#138](https://github.com/yupiik/tools-maven-plugin/pull/138) and [#141](https://github.com/yupiik/tools-maven-plugin/pull/141) are still open, and their readings go when they merge. Then I will ask for your review with the list of changes.

rolfedh:
@rmannibucau @fpapon, the rework is ready (rebased on master, HTML renderer commit moved out, #139 and #140 readings removed), but I will push it after [#144](https://github.com/yupiik/tools-maven-plugin/pull/144) merges. On current master, one test of this pull request fails, `escapedAttributeReferenceInABlockTitle`, because of an escaping bug from my [#140](https://github.com/yupiik/tools-maven-plugin/pull/140) that #144 fixes. With #144, all tests pass.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/html/AsciidoctorLikeHtmlRenderer.java
```diff
@@ -90,6 +92,8 @@
 public class AsciidoctorLikeHtmlRenderer implements Visitor<String> {
     protected final StringBuilder builder = new StringBuilder();
     protected final Configuration configuration;
+    protected final VisitorSibling sibling; // reads the model as the other renderers do
+    protected final VisitorState visitorState; // the section index shared with the other renderers, its document is state.document
```
rmannibucau, line 96: 🤔 isnt it fishy we have two states now? the +157LoC -142LoC doesnt reflect the target to share the state and sibling IMHO so it is probably wrongly abstracted, no?

rolfedh: @rmannibucau, yes. The renderer's own `State` stayed next to the new `VisitorState`, so the document and the footnotes are held twice. I am making `State` extend `VisitorState`: one state, the renderer delegates the document, the index and the footnotes to it and keeps only its HTML flags. It comes with the HTML renderer commit, in its own pull request after this one is reworked.

rolfedh:
@rmannibucau Thanks for the two reviews. The parsing point was right, and the parser fixes it produced, #136 to #140, were worth doing on their own.

One question before I push the rework, so I do not send you the wrong shape. The rework moves the HTML renderer commit into its own pull request, as I proposed on 14 September. That leaves `VisitorSibling` and `VisitorState` in `io.yupiik.asciidoc.renderer` with only the Markdown renderer using them, which is the case you said belongs in the markdown package. Would you rather have the HTML renderer commit back in here, the classes in the markdown package, or the shared package with the HTML renderer following next?

Three re-parses would stay: the `NOTE#id.role` shorthand, `%collapsible%open` as one key, and `kbd:[Ctrl,Shift]`. One parser pull request each as follow-ups, if that suits you?

rmannibucau:
@rolfedh ideally we get the shared state and sibling done right upfront since it will avoid PR and round trips _after_

thinking out loud: what if we avoid the reparsing and do it upfront in the parser? any case it can break (I think it just opens the syntax/is more tolerante but I don't see harmful cases).

rolfedh:
@rmannibucau  thank you for your thoughtful comments. I'll get back to this one later this week.

rolfedh:
@rmannibucau Reworked and pushed, following your comment of 21 September.

**The shared state and sibling, done in this pull request.** `AsciidoctorLikeHtmlRenderer.State` now extends `VisitorState`, so one state holds the document, the section index and the footnotes and the renderer keeps only its HTML flags. That answers your point about two states. The HTML renderer also reads through `VisitorSibling`, so both renderers use the abstraction and it stays in `io.yupiik.asciidoc.renderer`. `IdGenerator` moved out of `renderer.html` into `renderer`, with a deprecated delegate left behind, because the shared helper should not depend on the HTML package.

**The re-parsing, moved into the parser.** Two pull requests, both open with a green build: #156 reads the `#id`, `.role` and `%option` shorthands of a block style, and #157 keeps every key of `kbd:[Ctrl,Shift]`. When they merge I rebase this one and add a commit deleting the matching readings from `VisitorSibling`. I dropped the kbd commit that used to be here, so the two pull requests do not do the same work twice.

**Measured on the 282 quarkus.io guides**, each commit against the one before it:

| commit | guides whose HTML changes |
| --- | --- |
| a later footnote reusing an earlier id | 0 |
| the HTML renderer on `VisitorSibling` and `VisitorState` | 0, byte-identical |
| cross references | 23, only in links |
| image targets and alt texts | 4, only in 86 alt texts |
| options written in the style or grouped | 1 |
| `IdGenerator` moved | 0 |

The Markdown of all 282 guides is byte-identical across the whole series.

**What is still read in the renderer, and why.** The Markdown renderer still guesses where a paragraph ends inside a delimited admonition or a list item, in `segments`, `splitParagraphRun` and `splitBlockRun`. That is the family you flagged on 14 September, and it needs #141 so the parser gives the paragraphs; #141 has had no review since 16 September. One split also stays in `VisitorSibling.kbdKeys`, turning the bracket content into keys, because `Map<String, String>` cannot hold a list. #157 explains that one.

447 tests run and one fails, `escapedAttributeReferenceInABlockTitle`, which needs #144. I pushed rather than wait, since you approved #144 on 16 September.

asciidoc-java/src/test/java/io/yupiik/asciidoc/renderer/VisitorSiblingTest.java
```diff
@@ -0,0 +1,127 @@
...
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class VisitorSiblingTest { // the options come from the parser, so a change of the parser shows here
+    private final VisitorSibling sibling = new VisitorSibling();
+
+    @Test
+    void idAndStyleShorthands() {
+        final var note = ((OpenBlock) parse("""
+                [NOTE#note.important%collapsible]
```
rmannibucau: I would collapse the 3 asserts in an `assertIdShorthand` and test 

```
[NOTE#note.important%collapsible]
[NOTE.important#note%collapsible]
[NOTE,id=node,role=important%collapsible]
[NOTE,role=important,id=node%collapsible]
```

ensuring they all lead to the same

rmannibucau (changes requested):
a small test enhancement - also %collapsible%open can be neat to validate but not sure we want it there

and the point about double escaping of attributes with this code which was not answered

else ok for me

rolfedh:
#144 merged since I wrote that, so this is rebased on master `169314d` and the build is green: 473 tests, no failure. #156 and #157 are rebased too and both green. #156 needed one change beyond the rebase, since #154 brought in the same `addRole` helper and the two would have compiled twice; it now uses yours.

rolfedh:
**The double escaping point.** You were right, and it was only the link target. `VisitorSibling.substitute()` ran on `Link.url()`, which the parser had already substituted, and in doing so the parser had dropped the backslash of an escaped reference. So `\{name}` arrived as `{name}` and the second pass replaced it. `https://x.org/\{name}[label]` and `https://x.org/{name}[label]` both gave `value`, and the escape did nothing.

| source | model | Markdown before | Markdown now | HTML renderer |
| --- | --- | --- | --- | --- |
| `link:\{name}[label]` | `url={name}` | `[label](value)` | `[label]({name})` | `href="{name}"` |
| `https://x.org/\{name}[label]` | `url=https://x.org/{name}` | `...value` | `...{name}` | `href="...{name}"` |
| `image::\{name}.png[]` | `label=\{name}.png` | `{name}` | unchanged | unchanged |
| `image::a.png[link=\{name}]` | `opts={link=\{name}}` | `[a]({name})` | unchanged | unchanged |

A `Link` url is now used as the parser gives it, which is what the HTML renderer already did. The other `substitute()` calls stay, because the model still holds those values raw, as the last two rows show.

**The test helper.** The three asserts are now `assertNoteShorthand`, run on the same note written four ways: `[NOTE#note.important%collapsible]`, `[NOTE.important#note%collapsible]`, `[NOTE%collapsible,id=note,role=important]` and `[NOTE,id=note,role=important,%collapsible]`. All four give a NOTE admonition with the id `note` and the collapsible option.

Your third and fourth examples cannot join them, and I think that is correct rather than a gap: in `[NOTE,id=note,role=important%collapsible]` the `%collapsible` is part of the role value, giving the role `important%collapsible` and no option, and in `[NOTE,role=important,id=note%collapsible]` the id becomes `note%collapsible`. Asciidoctor reads the shorthand on the style, never on a named attribute. The test now asserts that difference rather than hiding it. Say if you read asciidoctor differently and I will change it.

**`%collapsible%open`** is already in that test, with `assertNull` on the id and both options asserted.

474 tests, all green. No guide writes an escaped reference in a link target, so the Markdown of the 282 guides is byte-identical.

rmannibucau:
for the test I think you are right even if I didnt run but point was to ensure it is parsed as expected and we dont extract it blindly (we had that phase for some feature in early times).

For the substitutes, did you investigate it was a parser laziness or expected? I tend to think it just a lack of earlysubstitution call

rolfedh:
Investigated. It is not a missing substitution: the parser already substitutes every macro target. `image::{name}.png[]` gives the label `value.png` and `xref:{name}[]` gives `value`.

What is missing is the unescape, and only on macros. #144, which was mine, added `unescapeAttributeReferences` at the two places that build a `Link`, `Parser.java:1959` and `1979`, and nowhere a macro label is built. On master today:

| source | model | master HTML |
| --- | --- | --- |
| `link:\{name}[l]` | `url={name}` | `<a href="{name}">l</a>` |
| `image::\{name}.png[]` | `label=\{name}.png` | `` |
| `xref:\{name}[]` | `label=\{name}` | `<a href="\{name}">\{name}</a>` |

So a link is right and a macro keeps the backslash in both the target and the visible text. The Markdown renderer of this pull request hides the difference, because `VisitorSibling.substitute()` removes the backslash, and that is the asymmetry you are pointing at.

I would fix it in the parser: unescape a macro label where it is built, as #144 does for a link. Every target is then ready to use, and `substitute()` on a target becomes unnecessary in the renderers. The guides write 32 escaped references, none of them in a macro target or a link url, so no guide output changes. Small pull request, shall I open it?

rolfedh:
Opened as #158. It unescapes the macro label where it is built, so `image::\{name}.png[]` and `xref:\{name}[]` stop writing the backslash into the target and the visible text, as `link:` already did.

One thing I measured while writing it, and said on #158 too: once it merges, `VisitorSibling.substitute()` on a macro target becomes a second pass over a value that is already unescaped, and the escape is lost again. With both applied, `image::\{name}.png[]` gives `value`. So when #158 merges I drop the target substitution here, at `imageTarget`, at the `video`/`audio` case and at the two `crossReference` call sites. Nothing on master depends on that, so #158 stands on its own.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/VisitorSibling.java
rolfedh: @rmannibucau You were right for a link target. Commit [`1ee0dd1`](https://github.com/yupiik/tools-maven-plugin/commit/1ee0dd17feadc3ed97337611f8d6bd518c075e08), pushed after your review, fixes it. An image target keeps the pass until [#158](https://github.com/yupiik/tools-maven-plugin/pull/158) merges.

On this branch `https://x.org/\{name}[label]` now gives `https://x.org/{name}` in both renderers, as asciidoctor does; before, the second pass wrote the attribute value.

The parser still hands an image target raw: the model label of `image::\{name}.png[]` is `\{name}.png`. The Markdown renderer writes `{name}`, which matches asciidoctor; the HTML renderer writes `alt="\{name}"`, which does not. #158 fixes that in the parser, and when it merges I drop the substitution here at `imageTarget`, at the video and audio case, and at the two `crossReference` call sites.

asciidoc-java/src/test/java/io/yupiik/asciidoc/renderer/VisitorSiblingTest.java
rolfedh: @rmannibucau Done in [`33cd702`](https://github.com/yupiik/tools-maven-plugin/commit/33cd7020db27887a54ed7107ff683837162bd00d): `assertNoteShorthand` runs the same note written four ways and asserts they give the same level, id and option.

Your first two lines give that block. Your last two cannot, so the test pins the difference rather than hiding it: a `%option` written against a `key=value` attribute is part of that value. `[NOTE,id=node,role=important%collapsible]` gives the role `important%collapsible`, and `[NOTE,role=important,id=node%collapsible]` gives the id `node%collapsible`; neither block has the option. Asciidoctor reads the shorthand on the style, never on a named attribute. I read your `id=node` as `id=note`.

rolfedh:
@rmannibucau Two commits are on top of the ones you reviewed, from a review pass I ran on the branch, and one more defect goes to a follow-up pull request.

[`8e0fdb5`](https://github.com/yupiik/tools-maven-plugin/commit/8e0fdb5cb8786f43dd6d7d2b2765e83e4d441d8a) reads a cross reference text as asciidoctor does: the `reftext` of the target, else its title, else the id between square brackets. The index held section titles only, so a link to any other anchored block showed the bare id. Of the 251 quarkus.io guides this parser renders, 15 change: 21 link texts become `[id]`, 5 become the title of the target.

[`7bd98f2`](https://github.com/yupiik/tools-maven-plugin/commit/7bd98f2ece096b6c3b6002685e6a1ec97ad5a485) collapses an example block only in Markdown, as the HTML renderer already did. No guide changes.

The third defect: a plain text run reaches the Markdown output unescaped, so `a*b*c` becomes italic on GitHub. It is in the known gaps now.

rmannibucau:
second one is fine, first one mainly has this substitute thing which can bite you anytime you use escaping and a small style question (escape)

rolfedh:
@rmannibucau Both substitute points are answered. The first commit was wrong, and [`c6b91c0`](https://github.com/yupiik/tools-maven-plugin/commit/c6b91c0c32af1cc239044eae2fa7e7a5a814ca2e) drops its second pass. For the image target you asked about, it is a lack of an early call, and [#158](https://github.com/yupiik/tools-maven-plugin/pull/158) is that call.

`VisitorSibling.referenceText()` called `substitute()` on the reftext and on the title of the target, which the parser had already substituted, [#140](https://github.com/yupiik/tools-maven-plugin/pull/140) for a block title and [#144](https://github.com/yupiik/tools-maven-plugin/pull/144) for the escape. The second pass undid the escape: with `:name: value` and a block titled `.My \{name} title`, a `<<t>>` link showed `My value title` where asciidoctor shows `My {name} title`. Both values are now read as the parser gives them, checked on every element type that carries a title. 479 tests, two of them new, and no guide changes.

For the image target, the parser substitutes the reference but keeps the backslash of an escaped one: `image::\{name}.png[]` gives the label `\{name}.png`. #158 makes the parser drop it, and once #158 merges the renderer call goes, as its description says.

Which `escape` do you mean? I would rather ask than guess between the `escape(text, characters)` helper of the Markdown renderer and the `escape()` call in `visitXref`.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/VisitorState.java
```diff
@@ -0,0 +1,384 @@
...
+            return text;
+        }
+    }
+
+    // walks the body once, separately from the state's own visit methods, so a renderer can delegate element visits to the state
+    private final class Index implements Visitor<Void> {
+        private int tocMacroLevels;
+
+        @Override
+        public ConditionalBlock.Context context() {
```
rmannibucau: neat: not needed I think

rmannibucau:
@rolfedh 

> return escape(referenceText != null ? referenceText : "[" + id + "]", "[]");

normally id are known safe so no need to make computed an output we know upfront until i'm missing a case

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/VisitorState.java
rolfedh: @rmannibucau Good question, I checked it, and it turns out to be needed. `Visitor.context()` has a default that returns `key -> null`, and `Index` implements `Visitor`, so inside `Index` a plain `context()` call resolves to that default instead of the state's context. I tried removing the override: `contextOverrideAppliesToTheIndexToo` fails, because the index then reads no `idprefix` and does not see `toc`, so its section ids and its conditional branch no longer match the renderer's. I could instead write `VisitorState.this.context()` at the six call sites, if you prefer that. And if you were pointing at the `Index` class itself rather than this method, tell me and I will look at that too.

rolfedh:
@rmannibucau Done in [`80a2ef8`](https://github.com/yupiik/tools-maven-plugin/commit/80a2ef821b6c44ad7c7abb0c3f5369809800ee5f): the bracket branch is written as `"\\[" + id + "\\]"` and only a reference text goes through `escape()`. The HTML renderer gets the same shape in `visitXref`, `'[' + reference.id() + ']'` without `escape()`. Tests and the 251 guides are unchanged.

rmannibucau (approved):

rolfedh:
@fpapon when you have a moment, could you merge these three? Romain approved each of them:

- https://github.com/yupiik/tools-maven-plugin/pull/133, this one, approved today at 13:29 UTC
- https://github.com/yupiik/tools-maven-plugin/pull/158, approved on 22 September
- https://github.com/yupiik/tools-maven-plugin/pull/159, approved today at 09:59 UTC

All three build green and merge cleanly on master. #157 is the only open pull request that overlaps #133, in the HTML renderer, and I will rebase it as soon as #133 is in. Thank you!

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/markdown/GithubFlavoredMarkdownRenderer.java
rolfedh: @rmannibucau Not any more. Since [`e689d62`](https://github.com/yupiik/tools-maven-plugin/commit/e689d62) the constructor builds the state, and `VisitorState` starts from an empty document, so `visitBody` without `visit()` works for a partial rendering.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/VisitorSibling.java
rolfedh: @rmannibucau Done in [`73ea32e`](https://github.com/yupiik/tools-maven-plugin/commit/73ea32e), as you wrote it.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/VisitorSibling.java
rolfedh: @rmannibucau Agreed, they are a stopgap. The parser gives the paragraphs of a list item or an admonition as sibling texts, so the renderer guesses the boundary. [`73ea32e`](https://github.com/yupiik/tools-maven-plugin/commit/73ea32e) moved them out of `VisitorSibling` into the Markdown renderer as `splitParagraphRun`, and [#141](https://github.com/yupiik/tools-maven-plugin/pull/141) makes the parser keep each paragraph, so once it merges the renderer visits the structure and the split goes.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/VisitorSibling.java
rolfedh: @rmannibucau Done in [`73ea32e`](https://github.com/yupiik/tools-maven-plugin/commit/73ea32e): `titleText` returns the stripped text as is, and creates the builder only at the first whitespace run it has to rewrite, copying per segment.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/VisitorSibling.java
rolfedh: @rmannibucau Done in [`73ea32e`](https://github.com/yupiik/tools-maven-plugin/commit/73ea32e).

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/VisitorSibling.java
rolfedh: @rmannibucau Done in [`73ea32e`](https://github.com/yupiik/tools-maven-plugin/commit/73ea32e): it is `kbdKeys`.

PR #136, opened 2026-09-14
[asciidoc] Strip the blanks around attribute names and values, as asciidoctor does

rolfedh:
`doParseOptions` kept the blank after a comma in the name of an attribute. So `[source, java, subs="attributes"]` gave the key `" subs"`, and the substitution was not applied. The same happened to `role`, `cols`, `options`, `width`, `window`, and to include attributes such as `tag` after `leveloffset=+1,`. As in asciidoctor's `AttributeList`, the blanks around a name, around `=` and around an unquoted value are now dropped, and the blanks inside quotes are kept.

This is the most common way to write these lists in the quarkus.io sources. On the 279 guides, the HTML of 186 guides changes. As in asciidoctor, code blocks get their `role`, such as the CLI, Maven and Gradle tabs (117 guides), images their `align` or `width` (21 guides), links their `window` (6 guides), `subs=attributes+` code blocks show the attribute values (17 guides, for example `{postgres-image}` becomes `docker.io/library/postgres:18`), and `quarkus-maven-plugin` gets the `leveloffset` of its includes.

The generated configuration tables (109 guides) get their `cols` too, so their cells are now parsed. Their rows still collapse into one cell, because the parser does not read the `a|` and `h|` cell specifiers yet; I plan to fix that next.


rmannibucau (approved):
curiosity: did you check we can remove some `strip()` call now it is preformatted?

rolfedh:
Yes, I checked. Two `strip()` calls can go: the ones in front of `doParseOptions` in `parseQuoteLikeOptions` and `mapIf`; all tests pass without them. The two in front of `parseOptions` must stay, since `parseOptions` compares prefixes such as `source,` before `doParseOptions` runs. The three on `citetitle` and `opts` would change a quoted value with blanks at its ends, which asciidoctor keeps. Should I remove the two in this PR?

rmannibucau:
if not too bothersome yes please

rolfedh:
Done in 43bbd01: I removed the two `strip()` calls in front of `doParseOptions`, in `parseQuoteLikeOptions` and `mapIf`.

PR #137, opened 2026-09-14
[asciidoc] Close a conditional block at endif::name[], and do not count ifdef::name[content] as an opening

rolfedh:
`readIfBlock` closed a conditional block only at `endif::[]`. A block closed with `endif::name[]` took the rest of the document: later sections became part of the block, so a false condition hid them too, and in the header a false condition lost the rest of the header and the whole body. It also counted the one-line form `ifdef::name[content]`, which has no `endif`, as an opening, so the block around it did not close either.

As in asciidoctor's reader, `endif::name[]` now closes the last block opened (the name compared without case), and only the block form of `ifdef` and `ifndef` opens a block. A name that does not match fails; asciidoctor logs an error there. None of the 54 named `endif` lines of the quarkus.io sources has a wrong name. The four directives are read with the same rule, a whole line, so an `endif` line with text after its brackets fails as malformed.

On the 279 guides, the HTML of 152 guides changes, mostly empty `endifblock` divs that disappear. The text changes in 10 guides: content that the block hid is back (621 lines in `security-authentication-mechanisms`, the Gradle tab of `gradle-tooling`), and in `datasource` a list is a list again, except the item inside the conditional block, which is a separate problem.


asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
```diff
@@ -1964,27 +1964,36 @@ private String earlyAttributeReplacement(final String value, final Function<Stri
...
+                if (openNames.isEmpty()) {
+                    break;
+                }
+            } else if (openNames.size() == 1 && ("else::[]".equals(stripped) || stripped.startsWith("elsif::"))) {
                 branches.add(current);
                 branchConditions.add(stripped);
                 current = new ArrayList<>();
                 continue;
+            } else if (next.startsWith("ifeval::")) {
+                openNames.add("");
```
rmannibucau: why not the same logic?

asciidoc-java/src/test/java/io/yupiik/asciidoc/parser/ParserTest.java
```diff
@@ -254,6 +254,26 @@ void parseHeaderWithConditionalBlocks() {
         }
     }
 
+    @Test
+    void parseHeaderWithNamedEndif() { // the named endif did not close the block, so the rest of the document was part of it
+        final var content = List.of("""
+                = Title
+                ifndef::env-github[]
+                :toc: left
+                endif::env-github[]
```
rmannibucau, line 263: what if line is `endif::env-github1[] endif::env-github2[]`?

rmannibucau (approved):
generally ok, just needs the ifeval check

asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
rolfedh: Right. Done in f09b61c: `ifdef`, `ifndef`, `ifeval` and `endif` lines now use the same check, a whole line `name::target[text]`. Before, `ifeval::[1 == 1] is text` inside a block opened a block, so the `endif` after it failed; asciidoctor reads that line as text. Tests: `ifevalInsideAConditionalBlock` and `ifevalWithTextAfterItsBracketsDoesNotOpenABlock`.

asciidoc-java/src/test/java/io/yupiik/asciidoc/parser/ParserTest.java
rolfedh: Asciidoctor reads it as `endif::env-github1[...]` with text in the brackets, logs "malformed preprocessor directive - text not permitted", and keeps the blocks open. Before f09b61c the parser failed with the mismatch error, which was confusing; now it fails with `Malformed preprocessor directive, text not permitted: 'endif::env-github1[] endif::env-github2[]'` (test `textAfterEndifFails`).

rolfedh:
> generally ok, just needs the ifeval check

Done in f09b61c. `readIfBlock` now checks an `ifeval` line with the same whole-line rule as `ifdef`, `ifndef` and `endif`. Two new tests are the first `ifeval` tests of the project: `ifevalInsideAConditionalBlock` and `ifevalWithTextAfterItsBracketsDoesNotOpenABlock`. Is this the check you meant?

asciidoc-java/src/test/java/io/yupiik/asciidoc/parser/ParserTest.java
rmannibucau: might be worth testing in another round ;)

PR #141, opened 2026-09-15
[asciidoc] Add the keep-paragraphs attribute to keep each paragraph of a group of blocks

rolfedh:
The parser replaced a paragraph of one text with its text, also in the blocks of an admonition block, a list item or a callout item, so a renderer could not tell two paragraphs from one. On master, these two list items give the same `Paragraph[Text(BOLD, "First"), Text("Second")]`:

```asciidoc
* **First**
+
Second
```

```asciidoc
* **First**Second
```

Now each paragraph of such a group stays a `Paragraph`, also when it has one child. The first item gives `Paragraph[Paragraph[Text], Paragraph[Text]]`, and the second item does not change. A paragraph alone, and a paragraph that holds a block such as a list, are unwrapped as before. The parser attribute `keep-paragraphs` set to `false`, in the document or in the `Parser` attributes, gives the model of the earlier versions. This is the flip asked below ("I'd flip it and do a 1.2.0 on my side"); the first version of this pull request had the attribute off by default.

Checked on the 279 quarkus.io guides:

- The HTML renderer writes each kept paragraph in its own `div.paragraph`, which changes 138 guides with the same visible text. With `:keep-paragraphs: false`, the HTML of every guide is the same as on master.
- On 15 September, with the same reading as an attribute: the parser keeps 493 paragraphs, and unwrapping them again gives the master model for every guide.

The rendering was also checked on 15 September on 260 small documents: 20 places where such a group can be (for example a nested list, a table cell, a description list, a quote, a sidebar or the preamble) times 13 kinds of kept paragraph (a role, an id, a title, a link, an inline image, an xref, a footnote, a hard line break, and more). With `false`, the HTML renderer, `ShowerRenderer` and `TocVisitor` give the same output as on master. By default, the table of contents does not change; in the HTML, the visible text, the ids and the roles do not change, and no invalid nesting is added; in 33 documents a `<p>` inside a `<p>` goes away. The 60 pages of the Yupiik OSS minisites do not change either (see the comment above).

The HTML renderer changes in one method, used by checklist items and callout items: they write a kept paragraph as its inline element with the paragraph options, as before. Without this method, a callout text with a role or an id lost it. Two things stay as on master: the paragraphs of a group are inside one outer `div.paragraph`, and a callout item writes its paragraphs as `<span>` elements.

The attribute is documented in a new "Parser attributes" table in `asciidoc-java.adoc`, together with `callout-mismatch`.


rmannibucau:
I don't get this one, we can get the same rendering keeping paragraph and it is semantically more correct no?

rolfedh:
I don't think a renderer can do it with the current model. Today these two list items give exactly the same model:

```
* **First**
+
Second
```

```
* **First**Second
```

Both are `Paragraph[Text(BOLD, "First"), Text("Second")]`. The first one is two paragraphs, the second one is one paragraph, so the renderer cannot choose. A `====` admonition with `**First**`, a blank line and `Second` has the same problem.

So the parser must give a different model for the group. I used `OpenBlock` because the parser already uses it for a group of blocks (`[NOTE]` on `--`, example blocks). Do you have another shape in mind?

rmannibucau:
> Today these two list items give exactly the same model

it comes from the optimization of the model (unwrap paragraph if not needed) but we can add a paragraph for the  first example around the text.
the note is that if you do that you will have to not break tests and renderer (html) but it looks more relevant


>  already uses it for a group of blocks ([NOTE] on --, example blocks)

because this is an open block and not a paragraph ;)

so overall issue is more than when a line contains only a formatted text and a continuation with other text we dont know they are paragraph (if the first line is not only the bold text then you have a paragraph in the list so it is easy to detect items are paragraphs)

i think enabling the parser to keep paragraphs is acceptable (likely a toggle in the configuration) but it should also be made friendly with the renderers somehow

rolfedh:
I force-pushed 4470907, which adds the toggle you proposed: the new parser attribute `keep-paragraphs`, `false` by default. It is read like `callout-mismatch`, from the document or from the `Parser` attributes.

With `true`, each paragraph in the blocks of an admonition block, a list item or a callout item stays a `Paragraph`, also when it has one text. `* **First**` with a `+` paragraph `Second` gives `Paragraph[Paragraph[Text], Paragraph[Text]]`, and `* **First**Second` still gives `Paragraph[Text, Text]`. A paragraph alone does not change.

For the renderers, I checked the 279 quarkus.io guides. Without the attribute, the HTML is the same as master. With it, the HTML renderer writes each paragraph in its own `div.paragraph` without a change, and 138 guides change with the same visible text. Only checklist items needed a small renderer change.

rmannibucau:
change by itself looks good, I would just give a review of all elements cause it can break more rendering but +1 to get it onboard

rolfedh:
@rmannibucau, I did the review of all elements and force-pushed d8f8b9e with one fix: with the attribute, a callout text lost its role and its id. A callout item now writes a kept paragraph as without the attribute, like a checklist item.

I checked 260 small documents: a group in 20 places (nested list, table cell, description list, quote, sidebar, preamble, ...) with 13 kinds of paragraph (role, id, title, link, image, xref, footnote, hard line break, ...). Without the attribute, the HTML, `ShowerRenderer` and `TocVisitor` output is the same as master. With it, the visible text, the ids and the roles do not change, and no invalid nesting is added. The description has the details.

rmannibucau:
@fpapon reversing the default has some impact but can also make sense, any opinion?

rolfedh:
@rmannibucau @fpapon, some data for the choice. I flipped the default locally: 3 existing `ParserTest` expectations change (a list item or callout text with an attached code block becomes `Paragraph[Text]` inside the group), and no existing HTML test changes. On the 279 quarkus.io guides, the HTML of 138 guides changes with the same visible text: the paragraphs of a group are written separately, as asciidoctor does. `:keep-paragraphs: false` gives back the old model. If you want the flip, I can push it to this pull request.

rmannibucau:
i know themes were adjusted to avoid spacing, think definition list too, a good test would be to diff the yupiik OSS minisites but not sure ill get time to do that this month (in particular the review).
breaking is fine but might need a 1.2.0 IMHO

rolfedh:
@rmannibucau, I diffed the minisites of `tools-maven-plugin`, `fusion`, `bundlebee`, `uship` and `yupiik-logging`, 60 pages together: no page changes with the attribute on. The attribute changes a list item, an admonition block or a callout item whose text is followed by a second paragraph, and these documents have none: their 7 `+` continuations attach a code block, a list or an admonition, and no admonition block holds two paragraphs. Description lists are not concerned.

Where the construct is used the difference shows. I rendered a page that uses it with the minisite theme: with the attribute off, "Install the extension." and the paragraph after the `+` are glued in one `<p>`; with it on they are two paragraphs, so the second takes the 16px bottom margin of `.page-content-body p`. An admonition block with two paragraphs behaves the same, and the callout list and the description list of that page do not change. The two renderings are side by side, in the theme, at https://claude.ai/artifact/GZkCqx7QGdoV63Vb35yvxp. In the 279 quarkus.io guides, 138 pages change.

The default stays off in this pull request, so nothing breaks. You wrote that breaking is fine but might need a 1.2.0: if you and @fpapon decide to reverse the default, tell me which version you mean and I will push the flip.

rmannibucau:
if there is nothing more crazy than that I'd flip it and do a 1.2.0 on my side

rolfedh:
@rmannibucau, done in 7340496: the default is flipped. Each paragraph of a group stays a `Paragraph`, and `:keep-paragraphs: false`, in the document or in the parser attributes, gives the earlier model. 338 tests pass; the three existing list tests with a `+` code block now expect the kept paragraph. The HTML of 138 quarkus.io guides changes, as measured before with the attribute on, and the 60 minisite pages do not. The description is updated. CI fails on `blockAnchor` only, the master failure that [#147](https://github.com/yupiik/tools-maven-plugin/pull/147) fixes. You do the 1.2.0 on your side, as you wrote.

rolfedh:
@rmannibucau @fpapon Rebased on master `169314d`: the conflict was in `ParserTest.java` only, where both sides added tests at the end of the file. Nothing else changed. 390 asciidoc-java tests and 35 minisite-core tests pass, and the description stays as it is.

Issue #142, opened 2026-09-15
[asciidoc] Parser and HTML renderer gaps found on the quarkus.io guides


rolfedh:
While working on the Markdown renderer (#133), I compared asciidoc-java with Asciidoctor 2.0.26 on the 279 quarkus.io guides. This is the list of gaps I found, checked on master `06bfa96`. I plan to fix the high and medium ones first, one pull request per gap, as for #136 to #141. Tell me if you prefer another order.

The counts come from the quarkus.io guides and the generated configuration documentation. The older issues #127 and #128 are not repeated here.

## Already in a pull request

| Pull request | Gap |
| --- | --- |
| #136 | A named attribute after a comma and a space keeps the space in its name, so `subs`, `role`, `cols` and include attributes written that way are ignored |
| #137 | A named `endif::name[]` never closes a conditional block, so the block swallows the rest of the document |
| #138 | A `[[id]]` line above a block is stored as its style: the id is lost and links to it break |
| #139 | Document attributes are copied into the options of a description list |
| #140 | Attribute references in block titles and block attribute lines are not substituted |
| #141 | Separate paragraphs are merged in an admonition block, a list item and a callout item |
| #143 | A cell specifier before `\|` (`a\|`, `h\|`) is not recognized, so the row collapses into one raw cell |

## High

| Gap | Occurrences |
| --- | --- |
| The HTML renderer drops the titles of admonitions, paragraphs and lists | 139 admonitions in 87 guides, 49 paragraphs in 17 guides, 16 lists in 8 guides |
| `ifdef`, `ifndef` and `endif` inside a code block are kept as code | 7,455 lines in 107 guides |
| `opts=optional` is ignored, so a missing optional include fails the document | 159 lines in 108 guides |
| A passthrough inside backticks, `` `+text+` `` or `` `+++text+++` ``, keeps its plus signs, and the attribute references inside it are substituted | 25,803 code spans in the HTML of 106 guides, from the generated configuration docs |

## Medium

| Gap | Occurrences |
| --- | --- |
| Inline markup in a block title, such as `` `code` `` or `*bold*`, is written as raw text | 111 code block titles in 25 guides, 17 other titles in 9 guides |
| In a table without `cols`, cell text is not parsed: `*bold*`, links and backticks stay raw | 2,206 cells in 247 tables of 114 guides, with #136 applied |
| The HTML renderer writes the `role` of a source block on `<code>`, not on the listing block | 550 source block lines with a role in 151 files |
[...]

PR #144, opened 2026-09-15
[asciidoc] Keep an escaped attribute reference, \{name} or {name\}, as asciidoctor does

rolfedh:
@rmannibucau, this fixes a regression of my [#140](https://github.com/yupiik/tools-maven-plugin/pull/140): since it substitutes block titles, `.Title with \{name}` is written `Title with \value`. The same early substitution already broke escaped references in text: `Text \{name}` gave `Text value`. [#133](https://github.com/yupiik/tools-maven-plugin/pull/133) needs this fix, because its Markdown test for an escaped title fails on master since #140.

As asciidoctor's `AttributeReferenceRx`, a reference with a backslash before `{` or before `}`, and a name made of word characters and `-`, is escaped and written `{name}`:

- The early substitution keeps it, and the inline parser writes it, as before, in text and list items.
- The parser unescapes it where nothing substitutes later: block titles, block attribute lines, code blocks with `subs=attributes+`, code spans and link targets.
- A title from a block attribute line is substituted once more, because asciidoctor substitutes a title when it writes the block: `title="A \{name} B {name}"` gives `A value B value`.

I checked each case against Asciidoctor 2.0.26. On the 279 quarkus.io guides, 5 guides change, each code span now written as asciidoctor writes it, such as `%X{mdc-key}` in `logging` and `{includes}/prerequisites.adoc` in `doc-create-tutorial`.

Not changed: a single-quoted attribute value still keeps its quotes, a separate entry of [#142](https://github.com/yupiik/tools-maven-plugin/issues/142).


asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
```diff
@@ -1809,8 +1813,8 @@ enclosingDocument, new Reader(List.of(label)),
                                                                 Stream.of(entry("role", (l.options().getOrDefault("role", "") + " inline-code").stripLeading())))
                                                         .collect(toMap(Map.Entry::getKey, Map.Entry::getValue))));
                             }
-                        } else {
-                            elements.add(new Code(content, Map.of(), true, List.of()));
+                        } else { // a +passthrough+ is kept as written, otherwise an escaped reference loses its backslash as in a text
```
rmannibucau, line 2011: i might misread but we are in backticks so cant be a passthrough 🤔

rmannibucau (approved):
might be worth a small bench after all the changes, we were insanely fast compared to asciidoctor - at least on the jvm - so hope we don't abuse of regex and loops (dont think but a side note before the release)

asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java
rolfedh: @rmannibucau, in asciidoctor it can: `+...+` inside backticks is a literal monospace. With `:name: value`, Asciidoctor 2.0.26 writes `` `+\{name}+` `` as `<code>\{name}</code>` and `` `\{name}` `` as `<code>{name}</code>`, so the backslash stays only inside the passthrough.

The parser still keeps the `+` signs in such a code span and substitutes `{name}` inside it. The generated configuration docs write every property that way (`` `+++quarkus.foo+++` ``), so on master the HTML of 106 guides has 25,803 code spans with plus signs. It is a separate gap, and I added it to [#142](https://github.com/yupiik/tools-maven-plugin/issues/142).

rolfedh:
@rmannibucau, the bench confirms your concern, and three cheap checks fix it: with them, master and #143 are faster than 1.2.16, with the same HTML for all 279 quarkus.io guides.

Parse and HTML rendering of the 247 guides that 1.2.16 also renders, median of 10 rounds after 5 warmup rounds:

| Build | Median | With the checks |
| --- | --- | --- |
| 1.2.16 | 1,811 ms | |
| master `fe4c370` | 3,252 ms | 1,442 ms |
| master + #143 | 4,674 ms | 1,306 ms |
| master + #138, #141 or #144 | same as master | |

Most of the time goes into three checks that run for every line or text, older than these pull requests: the email autolink regex in `flushText`, the ordered list regex in `parseLine`, and the stream that builds `NOTE: ` and the other labels in `parseAdmonition`. #143 parses about 13,000 more table cells, so it pays them more.

The checks look for an `@` before the email regex, for a `.` or `)` right before the first space before the ordered list regex, and for a `: ` where a label can end before the admonition stream. I can open them as a small pull request before the release.

rmannibucau:
yep we should stay fast, first iteration (cold) is interesting too cause this a the real life case (maven plugin or CLI) too, this one is algo + classes loaded (but think we are ok there, just maybe more regex compilations but should be okish)

fpapon:
@rolfedh can you resolve the conflicts please?

rolfedh:
@fpapon, done: I rebased on master in 4a6719d, and the conflict is gone. CI still fails on `blockAnchor`, which fails on master too since [#138](https://github.com/yupiik/tools-maven-plugin/pull/138) and [#146](https://github.com/yupiik/tools-maven-plugin/pull/146) merged. [#147](https://github.com/yupiik/tools-maven-plugin/pull/147) fixes that test.

rolfedh:
@rmannibucau, the first pass in a fresh JVM gives the same picture. Same 247 guides, 7 fresh JVMs per build, timed from just before the first parse; the files are read before the timing starts.

| Build | First pass, median (min to max of 7) | Warm, median of 10, re-measured |
| --- | --- | --- |
| 1.2.16 | 3,876 ms (3,492 to 5,056) | 1,714 ms |
| master `df30559`, with #143 | 6,344 ms (5,925 to 8,133) | 4,087 ms |
| master + the three checks | 3,556 ms (3,145 to 4,711) | 1,249 ms |

With the checks, the first pass is back in the range of 1.2.16. Master without them is above that range in all 7 runs. The checks add no regex: 8 lines, `indexOf` and `charAt` tests in front of the existing patterns.

rmannibucau (approved):

rolfedh:
@fpapon, the order for this one: [#147](https://github.com/yupiik/tools-maven-plugin/pull/147) merges first, then I rebase this on master, and the check can pass. Until then the `blockAnchor` failure stays, because it comes from master.

PR #146, opened 2026-09-15
[asciidoc] Fixes #49, write the id of a section on its title as asciidoctor does

### LINKED ISSUE #49: [asciidoc] Add anchor link and id on link instead of section div
Currently title don't have the anchor link added to them:

Should be:
```html
<div class="sect1">
<h2 id="_directory_structure"><a class="anchor" href="#_directory_structure"></a>Directory Structure</h2>
<div class="sectionbody">
...
</div>
```

Current:
```html
<div class="sect1" id="_directory_structure">
  <h2>Directory Structure</h2>
<div class="sectionbody">
...
 </div>
 </div>
```

rolfedh:
@rmannibucau, this fixes [#49](https://github.com/yupiik/tools-maven-plugin/issues/49): the HTML renderer now writes the id of a section on its title, `<h2 id="_directory_structure">`, as Asciidoctor 2.0.26 does, instead of on the section `div`.

- To keep the old HTML, use `Configuration.setSectionIdOnTitle(false)`, the `yupiik.asciidoc.sectionIdOnTitle` property of the `asciidoc` goal, or `--section-id-on-title false` on the launcher. `ShowerRenderer` always keeps the id on its `section`, because a slide is addressed by that id.
- `sectanchors` and `sectlinks` now also write the anchor and the link when the author sets the id, such as `[#custom]`. Before, they were written only for a generated id. Both attributes stay off by default, which answers your question in #49 about disabling them.
- Both attributes are read from the document, then from the configuration attributes, as `idprefix` is.

On the 279 quarkus.io guides, 252 guides change: in 4,359 sections the id moves from the `div` to the title, and each guide keeps the same ids. The minisite theme has no selector on these ids, and the minisite script now keeps the heading id instead of adding its own.

I updated the 22 test expectations that had the id on the `div`. New tests cover both values of the option and both attributes. A new table in `asciidoc-java.adoc` documents the options of `AsciidoctorLikeHtmlRenderer.Configuration`.


_documentation/src/main/minisite/content/asciidoc-java.adoc
```diff
@@ -83,3 +83,40 @@ The parser also supports a subset of CommonMark syntax for convenience:
 * Trailing double space at end of line → hard line break (`<br>`)
 
 These syntaxes are recognized alongside native Asciidoc syntax and can be mixed freely.
+
+== HTML renderer configuration
```
rmannibucau, line 87: saw you added a few sections like that in the doc but typically the tables must be generated using minisite preActions. For maven we already have a pluginXml parser/renderer, for the main we can do whatever works (maybe reusing fusion-cli and fusion-documentation or alike) but long story short I dont think we should start maintaining the doc manually for anything not design/architecture/tips - ie mainly configuration. Also we tend to prefer definitnion lists cause tables render super bad easily.

asciidoc-java/src/main/java/io/yupiik/asciidoc/renderer/html/AsciidoctorLikeHtmlRenderer.java
```diff
@@ -2157,6 +2175,16 @@ public Configuration setSectionTag(final String sectionTag) {
             return this;
         }
 
+        /**
+         * @param sectionIdOnTitle {@code true} writes the id of a section on its title tag ({@code h2}...) as asciidoctor does,
+         *                         {@code false} writes it on the section tag.
+         * @return this.
+         */
+        public Configuration setSectionIdOnTitle(final boolean sectionIdOnTitle) {
```
rmannibucau, line 2183: any real risk to always do it and not have a configuration?

rmannibucau (approved):
overall ok to me with 2 small notes:

- i dont think we need the new configuration
- we might want to ensure the doc configurations are generated and no more maintained manually whereever possible (maven for ex is clearly not something we want to write since we already have the tooling for it, cli needs some adjustment/rewrite to meet that requirement so no urgence)

PR #147, opened 2026-09-16
[asciidoc] Expect the section id on the heading in blockAnchor

rolfedh:
Master fails `AsciidoctorLikeHtmlRendererTest.blockAnchor` since [#146](https://github.com/yupiik/tools-maven-plugin/pull/146) and [#138](https://github.com/yupiik/tools-maven-plugin/pull/138), both mine, merged two minutes apart. #146 writes the id of a section on its title, and #138 added `blockAnchor` expecting that id on the `sect1` div. Asciidoctor 2.0.26 writes it on the title:

```html
<div class="sect1">
<h2 id="install">Install</h2>
```

This changes only that expectation. All 354 asciidoc-java tests pass.

PR #148, opened 2026-09-16
[asciidoc] Check an @, a list marker and an admonition label before the per-line regexes

rolfedh:
Three checks ran for every line or text of a document: the email autolink regex in `flushText`, the ordered list regex in `parseLine`, and the stream that builds `NOTE: ` and the other labels in `parseAdmonition`. A cheap check now comes first: an `@` in the text, a `.` or `)` right before the first space of the line, and a `: ` where an admonition label can end. No new regex, 8 lines.

Parse and HTML of the 247 quarkus.io guides that 1.2.16 also renders, measured today on master `df30559`:

| | 1.2.16 | master | master + this |
| --- | --- | --- | --- |
| warm, median of 10 rounds after 5 | 1,714 ms | 4,087 ms | 1,249 ms |
| first pass in a fresh JVM, median of 7 | 3,876 ms | 6,344 ms | 3,556 ms |

The HTML of all 279 guides is byte-identical. This follows the bench note on [#144](https://github.com/yupiik/tools-maven-plugin/pull/144).

PR #151, opened 2026-09-19
[asciidoc] Write the blocks of a table cell in a div and its text in a paragraph, as asciidoctor does

rolfedh:
@rmannibucau, this is the renderer fix announced in [#143](https://github.com/yupiik/tools-maven-plugin/pull/143): the renderer no longer writes the blocks of a table cell inside `<p class="tableblock">`. Text goes in that paragraph and blocks go in `<div class="content">`, as asciidoctor does.

On the 282 quarkus.io guides with their generated configuration tables, the cells with a block inside that paragraph go from 19,544 in 123 guides to 492 in 4 guides. Of the 24,671 cells whose tables line up with asciidoctor 2.0.26, 22,756 now have the same wrapper as in asciidoctor, against 11,657 on master. The visible text and the ids of every guide do not change.

The model does not keep the style of a cell, so the renderer decides from the content: inline content goes in one `<p class="tableblock">`, several paragraphs in one each, any other content in `<div class="content">`, and an empty cell gets nothing. A header row cell holds its text alone. The span, the alignment and the header role are written on the `td` or `th` only, so a link in an `h|` cell no longer gets `class="header"`. `writeTableCell` keeps its signature: the header row is a flag in the renderer state.

Four test expectations lose the extra wrappers, and `tableCellContent` and `tableHeaderRowCells` are new. `blockAnchor` fails as on master; [#147](https://github.com/yupiik/tools-maven-plugin/pull/147) and [#149](https://github.com/yupiik/tools-maven-plugin/pull/149) both fix it.

Not fixed here:

- An `a|` cell of text only gets `<p class="tableblock">`, because the parser does not record the `a` style: 994 of those 24,671 cells.
- The 492 cells come from `visitText`, which writes `[[id]]text` and `[.role]#*text*#` as a `<p>`, also outside tables.
- Found while testing, from [#143](https://github.com/yupiik/tools-maven-plugin/pull/143): a header cell whose text has a role, such as `h|[.header-title]##Configuration property##`, becomes a `td`, because `merge()` keeps the role of the text over the header role: 504 cells in 335 generated files.


rmannibucau (approved):

rolfedh:
@fpapon, the build passes on all three now: [#144](https://github.com/yupiik/tools-maven-plugin/pull/144), [#148](https://github.com/yupiik/tools-maven-plugin/pull/148) and [#151](https://github.com/yupiik/tools-maven-plugin/pull/151).

They were failing on master's `blockAnchor` test, which your merge of [#147](https://github.com/yupiik/tools-maven-plugin/pull/147) fixed. I rebased each branch onto `daf3176` this evening. The rebase changed no content, each approval survived, and the three builds pass.

Thanks for merging #147 and #149.

